PR #1, opened 2016-12-12
Fixes boxing - Blaze global style

hopkins-tk:
Introduces & uses **generics.shared.scss** which is based on **~blaze/scss/generics.global.scss** with small tweak to be working in shadow DOM.

### DIFF
--- a/packages/alert/Alert.scss
+++ b/packages/alert/Alert.scss
@@ -1,2 +1,3 @@
+@import "../generics.shared";
 @import '~blaze/scss/components.alerts';
 @import '~blaze/scss/components.buttons';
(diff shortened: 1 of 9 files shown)


hopkins-tk (commented):
This new version with **:host** is not working in Firefox - maybe Polyfill is missing.

devopsmariocom:
AFAIK :host only worked with vanilla WC

CC: @hotell

hopkins-tk:
So do you think I should go with `*` and `* *` version?

hopkins-tk:
` *` and `* *` version works in IE as well, so I changed it back to it.

devopsmariocom (approved):
I don't like * * approach but seems that it's working solution for now so let's stick with that and we can replace it any time later

PR #3, opened 2016-12-12
feat(modal): modal accessibility improved

hopkins-tk:
Improves accessibility:
- ARIA roles introduced
- prevention of focus lost
- returning focus to the element that was focused when the modal was opened

### DIFF
--- a/packages/modal/Modal.tsx
+++ b/packages/modal/Modal.tsx
@@ -99,7 +99,7 @@ export class Modal extends Component<ModalProps> {
 
   renderedCallback() {
     if ( this.isOpen ) {
-      this.lastActiveElement = document.activeElement as HTMLElement;
+      this.lastActiveElement = this.deepActiveElement();
 
       this.focusModal();
 
@@ -112,6 +112,16 @@ export class Modal extends Component<ModalProps> {
       }
     }
   }
+
+  private deepActiveElement(): HTMLElement {
+    let activeElement = document.activeElement;
+    while (activeElement && activeElement.shadowRoot && activeElement.shadowRoot.activeElement) {
+      activeElement = activeElement.shadowRoot.activeElement;
+    }
+
+    return activeElement as HTMLElement;
+  }
+
 }
 
 customElements.define( Modal.is, Modal )
--- a/packages/modal/Modal.tsx
+++ b/packages/modal/Modal.tsx
@@ -5,6 +5,7 @@ import { Card } from '../card/Card';
 
 interface ModalProps extends JSX.HTMLProps<HTMLElement | any> {
   isOpen?: boolean,
+  closeTitle?: string,
   onModalClose?: Function,
 }
 export class Modal extends Component<ModalProps> {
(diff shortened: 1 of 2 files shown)


hopkins-tk:
I finally found how to get focused element from Shadow DOM
- see https://developers.google.com/web/fundamentals/getting-started/primers/shadowdom

packages/modal/Modal.tsx
```diff
@@ -14,62 +15,113 @@ export class Modal extends Component<ModalProps> {
...
+
+      if (this.lastActiveElement) {
+        this.lastActiveElement.focus();
+      }
     }
   }
+
+  private deepActiveElement(): HTMLElement {
+    let activeElement = document.activeElement;
+    while (activeElement && activeElement.shadowRoot && activeElement.shadowRoot.activeElement) {
```
devopsmariocom, line 118: I have perf concerns about this loop, I get the point what is solve but I'm not sure if you really need deep nested focus

hopkins-tk: You just need to find the active element (the one that is **actually** focused).

Otherwise first WebComponent element will be taken and returned and you would have to make sure, to handle focus of that element to focus whatever needs to be focused.

There might be multiple focus-able items within the component, so you might not be able to know which is correct and should be focused when the element received focus.

hopkins-tk: @Hotell what do you think?

Hotell: huh I don't like this while at all folks

Hotell: in edge cases like this, we should always reach out to Polymer stuff ( yes they stack is a mess, BUT they already invented all of these A11y patterns and so on ) , so check out this https://github.com/PolymerElements/iron-overlay-behavior/blob/master/iron-overlay-manager.html#L68 I guess it's similar to @hopkins-tk solution

Hotell: demo https://elements.polymer-project.org/elements/iron-overlay-behavior?view=demo:demo/index.html&active=Polymer.IronOverlayBehavior

Hotell: also I highly recommend skimming through this source https://github.com/PolymerElements/iron-overlay-behavior/blob/2.0-preview/iron-overlay-behavior.html

hopkins-tk: Yes it is basically the same - they are looping till they find the deepest **activeElement**.

We could extract the logic out a bit - similarly as Polymer has it (`Polymer.dom`), but the need for deep search will stay, if we would like to support focus-return feature...

Hotell: AFAIK Polymer.dom is gonna be deprecated in Polymer 2.0 

> Shadow DOM v1
Polymer 2.0 supports shadow DOM v1. 
The shady DOM shim that was part of Polymer 1.x has been factored out of Polymer and added to the webcomponents-lite.js polyfill bundle, along with the related shim for CSS Custom Properties. 

> **This new version of shady DOM no longer exposes an alternative (Polymer.dom) API but instead patches the native DOM API, so 2.0 users can use the native DOM APIS directly.**

devopsmariocom (approved):

Issue #6, opened 2016-12-13
Create tree components


davidrus:
Looks like http://blazecss.com/components/trees/ and have collabsible logic.

PR #7, opened 2016-12-13
Tree WC component

davidrus:
This is implementation of #6

### DIFF
--- a/packages/index.demo.ts
+++ b/packages/index.demo.ts
@@ -6,3 +6,4 @@ import './input/Input.demo';
 import './modal/Modal.demo';
 import './toggle/Toggle.demo';
 import './tooltip/Tooltip.demo';
+import './tree/Tree.demo';
--- a/packages/tree/Tree.scss
+++ b/packages/tree/Tree.scss
@@ -1,2 +1,5 @@
 @import "~blaze/scss/components.trees";
-@import "~blaze/scss/components.links";
+
+ul {
+  padding-left: 1em !important;
+}
--- a/packages/tree/Tree.scss
+++ b/packages/tree/Tree.scss
@@ -1,0 +1,2 @@
+@import "~blaze/scss/components.trees";
+@import "~blaze/scss/components.links";
--- a/packages/tree/package.json
+++ b/packages/tree/package.json
@@ -1,0 +1,11 @@
+{
+  "name": "bl-tree",
+  "version": "1.0.0",
+  "main": "dist/main.js",
+  "license": "MIT",
+  "scripts": {
+    "start": "../../node_modules/.bin/webpack-dev-server  --config ../../webpack.config.js --env.dev --hot",
+    "build": "../../node_modules/.bin/rimraf dist && ../../node_modules/.bin/webpack --config ../../webpack.config.js --env.element=tooltip",
+    "prepublish": "npm run build"
+  }
+}
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "input",
     "modal",
     "toggle",
-    "tooltip"
+    "tooltip",
+    "tree"
   ]
 }
--- a/packages/index.ts
+++ b/packages/index.ts
@@ -6,3 +6,4 @@ import './input/Input';
 import './modal/Modal';
 import './toggle/Toggle';
 import './tooltip/Tooltip';
+import './tree/Tree';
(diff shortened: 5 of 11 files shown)


packages/index.demo.ts
```diff
@@ -6,3 +6,4 @@ import './input/Input.demo';
 import './modal/Modal.demo';
 import './toggle/Toggle.demo';
 import './tooltip/Tooltip.demo';
+import './tree/Tree.demo';
```
devopsmariocom, line 9: please add to packages/index.ts as well

packages/tree/Tree.scss
```diff
@@ -0,0 +1,5 @@
+@import "~blaze/scss/components.trees";
+
+ul {
+  padding-left: 1em !important;
```
devopsmariocom, line 4: where this come from? it should be either part of global css which is currently under PR #1 at least put some comment here. thx

packages/tree/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "bl-tree",
+  "version": "1.0.0",
+  "main": "dist/main.js",
+  "license": "MIT",
+  "scripts": {
+    "start": "../../node_modules/.bin/webpack-dev-server  --config ../../webpack.config.js --env.dev --hot",
+    "build": "../../node_modules/.bin/rimraf dist && ../../node_modules/.bin/webpack --config ../../webpack.config.js --env.element=tooltip",
```
devopsmariocom, line 8: please replace `tooltip` with `tree`

devopsmariocom (changes requested):
I'm looking at https://beta.webcomponents.org/element/PolymerElements/paper-menu I'm not sure that their approach is good UX but what I like is more semantic way. Is there any way how to have it e.g. 

```
<Tree>
  <TreeItem>
    Directory 1
    <Tree>
      <TreeItem>File 1</TreeItem>
      <TreeItem>File 2</TreeItem>
    </Tree>
  </TreeItem>
</Tree>
```

hopkins-tk:
I agree with @elmariofredo regarding the suggested way.

From my point of view, it would be also nicer.

Wouldn't using of 
```
<slot></slot>
```

instead of
```
<slot></slot>
{ isOpen ? <slot name="subItems" /> : null }
```

help?

elemementor:
guys I'm happy you are pushing my idea forward, as I said to @elmariofredo I'm currently busy so you are on your own.

packages/tree/Tree.scss
davidrus: I've added this, because shadow-dom does not match css selectors like ".c-tree .c-tree" which original css have [here](https://github.com/BlazeCSS/Blaze/blob/master/scss/components.trees.scss#L10) but I suggest to replace selector with class '.c-tree' what do you think? @elmariofredo

packages/tree/package.json
davidrus: Replaced

packages/index.demo.ts
davidrus: Yeah, I forgot, thanks 👍

davidrus:
@hopkins-tk we need to have condition for this slot because we are removing subitems from DOM when menu is collapsed

@elmariofredo do you think that we should have someting like this? 
```
<Tree>

  <TreeSubmenu>
    <TreeItem>Directory 1</TreeItem>
    <Tree>
      <TreeItem>File 1</TreeItem>
      <TreeItem>File 2</TreeItem>
    </Tree>
  </TreeSubmenu>

  <TreeItem>File Root 1</TreeItem>

</Tree>
```

packages/tree/Tree.scss
devopsmariocom: could we simply use `bl-tree` selector?

davidrus: No, because when someone (like us in demo) use jsx instead of html web element, this selector does not match.

devopsmariocom:
@davidrus well I would rather have my proposed structure. I did post link to polymer to have another opinion at the beginning of this PR so we can compare our decisions with something. For me their API is ugly and consist too much boilerplate.

I would try my proposal first and then let real world development to bring breaking changes later on. We shouldn't be afraid of doing elements even when we know that their API UX might not be perfect that's why we have each component separately and semver ;)

davidrus:
OK, I will try to do it now...

davidrus:
I removed ```hasSubItem``` flag, but have no idea how to remove ```<slot name="subItems" />``` and keep removing from DOM. 

So for now I suggest keep it as is. OK? @elmariofredo

packages/index.demo.ts
devopsmariocom: np we should generate these files anyway ;)

elemementor (approved):

devopsmariocom (approved):

PR #11, opened 2016-12-16
Bubble element

### LINKED ISSUE #9: Bubbles
**Bubbles** are like **Tooltips**, but support HTML within.

http://blazecss.com/components/bubbles/

hopkins-tk:
Fixes #9

### DIFF
--- a/packages/bubble/Bubble.tsx
+++ b/packages/bubble/Bubble.tsx
@@ -14,7 +14,8 @@ type BubbleTypesType = typeof BubbleTypes;
 // public
 interface BubbleProps extends JSX.HTMLProps<HTMLElement | any> {
   type?: keyof BubbleTypesType,
-  isDisplayed?: boolean
+  isDisplayed?: boolean,
+  disableAutoShowHide?: boolean
 }
 
 export class Bubble extends Component<BubbleProps> {
@@ -27,19 +28,27 @@ export class Bubble extends Component<BubbleProps> {
       }),
       isDisplayed: prop.boolean({
         attribute: true
+      }),
+      disableAutoShowHide: prop.boolean({
+        attribute: true
       })
     }
   }
 
   type = 'right';
   isDisplayed = false;
+  disableAutoShowHide = false;
 
   private handleMouseOver() {
-    this.isDisplayed = true;
+    if ( !this.disableAutoShowHide ) {
+      this.isDisplayed = true;
+    }
   }
 
   private handleMouseLeave() {
-    this.isDisplayed = false;
+    if ( !this.disableAutoShowHide ) {
+      this.isDisplayed = false;
+    }
   }
 
   connectedCallback(){
@@ -62,7 +71,7 @@ export class Bubble extends Component<BubbleProps> {
     return [
       <style>{styles}</style>,
       isDisplayed
-        ? <div className={"bubble-wrapper"} onMouseleave={this.handleMouseLeave}>
+        ? <div tabIndex={0} className={"bubble-wrapper"} onMouseleave={this.handleMouseLeave} onBlur={this.handleMouseLeave}>
             <slot></slot>
             <div
               className={className}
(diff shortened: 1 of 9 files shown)


packages/bubble/Bubble.tsx
```diff
@@ -0,0 +1,90 @@
...
+        'c-bubble--bottom': type === BubbleTypes.bottom,
+      }
+    );
+
+    return [
+      <style>{styles}</style>,
+      isDisplayed
+        ? <div tabIndex={0} className={"bubble-wrapper"} onMouseleave={this.handleMouseLeave} onBlur={this.handleMouseLeave}>
+            <slot name={"handle"}></slot>
+            <div
```
davidrus: Just style: Merge these 3 lines into 1

packages/bubble/Bubble.tsx
```diff
@@ -0,0 +1,90 @@
...
+              className={className}
+            >
+              <slot></slot>
+            </div>
+          </div>
+        : <div tabIndex={0} className={"bubble-wrapper"} onMouseover={this.handleMouseOver} onFocus={this.handleMouseOver}>
+            <slot name={"handle"}></slot>
```
davidrus: Why the name is expression? I think it will be better to use: ```<slot name="handle" />```

packages/bubble/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "bl-bubble",
+  "version": "1.0.1",
```
davidrus: Is this first version of bubble? If yes, please change version to ```1.0.0```

packages/bubble/Bubble.tsx
```diff
@@ -0,0 +1,90 @@
...
+        attribute: true
+      })
+    }
+  }
+
+  type = 'right';
+  isDisplayed = false;
+  disableAutoShowHide = false;
+
+  private handleMouseOver() {
```
davidrus, line 38: Please merge handleMouseOver and handleMouseLeave into one function, something like:
```
handleMouse(){
  if ( !this.disableAutoShowHide ) {
       this.isDisplayed = !this.isDisplayed;
   }
}
```
What do you think?

packages/bubble/Bubble.tsx
```diff
@@ -0,0 +1,90 @@
...
+  disableAutoShowHide?: boolean
+}
+
+export class Bubble extends Component<BubbleProps> {
+  _props: BubbleProps;
+  static get is() { return 'bl-bubble' }
+  static get props() {
+    return {
+      type: prop.string({
+        attribute: true
```
davidrus: Do we need reflect ```type``` into attribute? We dont change it in lifecycle, or yes? The same for ```isDisplayed``` and ```disableAutoShowHide```

davidrus (changes requested):
Please take a look at these small proposal in your code.

packages/bubble/Bubble.tsx
hopkins-tk: Good point - I will do that for:
- `type`
- `disableAutoShowHide`

But for `isDisplayed` I think it is something we might want.

packages/bubble/Bubble.tsx
hopkins-tk: I think this might lead to an unpredictable behavior in some cases as - there are two sources:
- mouse
- focus

If we merge it it would might got displayed based on focus and then hidden based on moseover...

packages/bubble/Bubble.tsx
hopkins-tk: Good point.

packages/bubble/package.json
hopkins-tk: Fixed.

packages/bubble/Bubble.tsx
hopkins-tk: Fixed.

packages/bubble/Bubble.tsx
davidrus: Good point, keep it as is.

davidrus (approved):

PR #19, opened 2016-12-19
Breadcrumb element

### LINKED ISSUE #15: Breadcrumbs
http://blazecss.com/components/breadcrumbs/

hopkins-tk:
Introduces breadcrumb element which can take several items and display them.
- it is possible to add **href** to the item
- it is possible to make some item active (selected)

Fixes #15

### DIFF
--- a/packages/breadcrumb/Breadcrumb-item.tsx
+++ b/packages/breadcrumb/Breadcrumb-item.tsx
@@ -4,7 +4,6 @@ import { css } from '../../utils/css';
 
 interface BreadcrumbItemProps extends JSX.HTMLProps<HTMLElement | any> {
   isActive?: boolean,
-  isLast?: boolean,
   href?: string
 }
 export class BreadcrumbItem extends Component<BreadcrumbItemProps> {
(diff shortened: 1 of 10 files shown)


packages/breadcrumb/Breadcrumb-item.tsx
```diff
@@ -0,0 +1,56 @@
...
+  _props: BreadcrumbItemProps;
+
+  static get is() { return 'bl-breadcrumb-item' }
+
+  static get props() {
+    return {
+      isActive: prop.boolean({
+        attribute: true
+      }),
+      isLast: prop.boolean(),
```
davidrus, line 19: please add ```attribute: true``` to keep functionality in not jsx elements

davidrus (changes requested):
Just thinking: What do you say to use ```<Link>``` element instead of href attribute?

davidrus (approved):

packages/breadcrumb/Breadcrumb-item.tsx
hopkins-tk: Good point I have created issue #22.

PR #20, opened 2016-12-20
Add nav element

### LINKED ISSUE #10: Navs
Navs is horizontal, or vertical main navigation bar with options from: http://blazecss.com/components/navs/

davidrus:
fixes: #10

### DIFF
--- a/packages/nav/Nav-item.tsx
+++ b/packages/nav/Nav-item.tsx
@@ -49,7 +49,6 @@ export class NavItem extends Component<NavItemProps> {
       {
         'c-nav__item--active': active,
         'c-nav__item--right': right,
-        'c-nav__content': type === NavItemTypes.content,
         'c-nav__item': type !== NavItemTypes.content,
         'c-nav__item--success': type === NavItemTypes.success,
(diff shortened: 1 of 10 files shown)


packages/nav/Nav-item.tsx
```diff
@@ -0,0 +1,70 @@
...
+  right: boolean;
+  inline: boolean;
+
+  renderCallback() {
+    const { type, active, right, inline } = this;
+    const className = css(
+      {
+        'c-nav__item--active': active,
+        'c-nav__item--right': right,
+        'c-nav__content': type === NavItemTypes.content,
```
hopkins-tk: It seams a bit weird to have **content** type for item...

It can be used for images as noted here: http://blazecss.com/components/navs/
But I would consider having it as another element - **NavContent** or something...

davidrus: I removed this type, so we will create navcontent wc in the future

davidrus:
Ready to re-review

hopkins-tk:
Great - so you can merge it ;-)

PR #23, opened 2016-12-20
Introduce drawer

### LINKED ISSUE #21: Drawer component
Add drawer component like http://blazecss.com/objects/drawers/
API should have:
- positioning: top, bottom, left, right
- dismissable: visible, or not

davidrus:
resolve #21

### DIFF
--- a/packages/drawer/package.json
+++ b/packages/drawer/package.json
@@ -1,0 +1,12 @@
+{
+  "name": "bl-drawer",
+  "version": "1.0.1",
+  "main": "dist/main.js",
+  "license": "MIT",
+  "scripts": {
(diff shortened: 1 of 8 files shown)


packages/drawer/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "bl-drawer",
+  "version": "1.0.1",
```
hopkins-tk: Should be **1.0.0**

hopkins-tk (changes requested):

packages/drawer/package.json
davidrus: thanks

hopkins-tk (approved):

PR #24, opened 2016-12-20
Introduce Link element

### LINKED ISSUE #22: Link element
Element representing link should be created.

It should be fairly simple:
```
<a class="c-link" href={href}>
    <slot />
</a>
```

davidrus:
resolve #22

### DIFF
--- a/packages/link/Link.tsx
+++ b/packages/link/Link.tsx
@@ -22,7 +22,7 @@ interface LinkProps extends JSX.HTMLProps<Link | any> {
   download?: string,
   hreflang?: string,
   rel?: string,
-  target?: keyof typeof LinkTargets,
+  target?: keyof typeof LinkTargets | string,
   type?: string,
   color?: keyof typeof LinkColors,
 }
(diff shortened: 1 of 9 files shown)


packages/link/Link.tsx
```diff
@@ -0,0 +1,103 @@
...
+  success: 'success',
+  error: 'error',
+};
+
+interface LinkProps extends JSX.HTMLProps<Link | any> {
+  href?: string,
+  download?: string,
+  hreflang?: string,
+  rel?: string,
+  target?: keyof typeof LinkTargets,
```
hopkins-tk: Target can be also any _frame name_ - so any string fulfilling certain criteria...

hopkins-tk (changes requested):

packages/link/Link.tsx
davidrus: Thanks, thats right

hopkins-tk (approved):

PR #26, opened 2016-12-21
Collapsible & accordion elements

### LINKED ISSUE #14: Accordions
http://blazecss.com/components/accordions/

hopkins-tk:
Introduces **Collapsible** & **Accordion** elements.

Fixes #14


packages/collapsible/Collapsible.tsx
```diff
@@ -0,0 +1,99 @@
...
+      isNotStandAlone: prop.boolean(),
+      isLast: prop.boolean()
+    }
+  }
+
+  isOpened: boolean;
+
+  // internal property to be used from other components, when there is collection of collapsible items to indicate
+  // this is the last one in the collection
+  isLast: boolean;
```
davidrus, line 36: Maybe we should use prefix `_` for internal properties. The same is for ```isNotStandAlone```

packages/accordion/Accordion.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  connectedCallback(){
+    super.connectedCallback();
+    this.handleStateChanged = this.handleStateChanged.bind(this);
+  }
+
+  renderCallback() {
+    const collapsibleItems = this.getElementsByTagName('bl-collapsible');
+
+    for(let i = 0; i < collapsibleItems.length; i++) {
```
davidrus: I prefer to save length into separate const, its just for performance.

davidrus (changes requested):

davidrus (approved):

packages/collapsible/Collapsible.tsx
hopkins-tk: As it is not really private, but like library internal, I would not suggest using `_`.

PR #37, opened 2016-12-27
Removal of all _props as generics are used

hopkins-tk:
- all **_props** has been removed from all existing components as generics are being used there

@Hotell is it OK to proceed with this now or should we wait for something?

### DIFF
--- a/packages/alert/Alert.tsx
+++ b/packages/alert/Alert.tsx
@@ -12,7 +12,6 @@ interface AlertProps extends JSX.HTMLProps<HTMLElement | any> {
 }
 
 export class Alert extends Component<AlertProps> {
-  _props: AlertProps;
   static get is() { return 'bl-alert' }
   static get props() {
     return {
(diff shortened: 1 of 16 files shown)


Hotell:
👋 yup, `_props` are going to be internal "hack" for TSX when this will be merged https://github.com/skatejs/skatejs/pull/952

we are overriding `Component` definition type somehow?

devopsmariocom:
https://github.com/skatejs/skatejs/pull/952 is already merged so what is the blocker?

Hotell:
wot, I'm working on this with new TS, there are other issues, not just removal of props, which I will address in separate Pull ( which will have a lot conflicts I guess :( ).
So should we close this or?

Hotell:
and this is rather related to this one https://github.com/skatejs/skatejs/pull/1012
that previous PR was initial step but not without problems

devopsmariocom:
I would close it as soon as you will have at least WIP PR here... what do you think?

Hotell:
sounds good! thx

Hotell:
Closed in favor of https://github.com/wc-catalogue/blaze-elements/pull/130

PR #43, opened 2016-12-28
fix(Nav.demo): use https for image

devopsmariocom:
to have proper https badge in GCh


hopkins-tk (approved):
Protocol relative (//some.domain/...) might be better pattern in the future.

PR #54, opened 2016-12-29
Fix of demo file name generation

hopkins-tk:
I would prefer this easy fix to be used to fix issues with current Pull Requests and than once we have the elements merged, we could do the bulk change as we have discussed it.


utils/generate-packages-list.js
```diff
@@ -15,5 +15,5 @@ fs.writeFileSync( './packages/index.ts', indexFile + "\n" );
 fs.writeFileSync( './packages/index.demo.ts', indexDemoFile + "\n" );
 
 function capitalizeFirstLetter( string ) {
```
devopsmariocom: please fix naming :)

devopsmariocom (approved):

PR #57, opened 2016-12-30
Refactoring of typography module

hopkins-tk:
All typography related elements moved to single module.


packages/typography/Code.scss
```diff
@@ -0,0 +1,2 @@
+@import "../generics.shared";
+@import "../../node_modules/blaze/scss/components.typography";
```
devopsmariocom: please use `~` instead of `../../node_modules/`

packages/typography/index.demo.tsx
```diff
@@ -0,0 +1,36 @@
+import { h, Component } from 'skatejs';
```
devopsmariocom, line 1: I guess we can remove demo files as well and put all demos directly here to shrink amount of files in this module... what do you think?

devopsmariocom (changes requested):

packages/typography/index.demo.tsx
hopkins-tk: I have tried that at first, but it was huge file, which was not very readable, so I restored the demo files.

devopsmariocom: yeah I meant single demo component with all examples inside, not multiple demo components within single file if that is what you mean

hopkins-tk: Done

devopsmariocom (approved):

hopkins-tk:
Closes #92

PR #96, opened 2017-01-02
chore(test): remove default set from CI command

devopsmariocom:


Hotell:
I have no rights. Shame

wct.conf.js
```diff
@@ -12,8 +13,8 @@ const config = {
           "version": "14.14393"
         }, {
           "browserName": "internet explorer",
-          "platform": "Windows 10",
-          "version": "11.103"
+          "platform": "Windows 7",
```
Hotell: why win 7 ?

wct.conf.js
```diff
@@ -1,6 +1,7 @@
 const packageJson = require( './package.json' );
 
 const config = {
+  testTimeout: 200 * 1000,
```
Hotell: what is this magic number? set in in constant with appropriate name pls

Hotell (changes requested):

Hotell (approved):

PR #97, opened 2017-01-02
feat(tag): add component

davidrus:
Basic tag component, after that I will create component container which allows to use tags as here: http://blazecss.com/components/tags/


packages/tag/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "bl-tag",
+  "version": "1.0.1",
```
hopkins-tk: :)

hopkins-tk (approved):

PR #113, opened 2017-01-05
fix(input): emit input event

devopsmariocom:


packages/input/Input.tsx
```diff
@@ -57,14 +57,16 @@ export class Input extends Component<InputProps> {
 
   inputElement: HTMLInputElement;
 
-  private provideValue(event: Event) {
+  private delegateEventToParent( event: Event ) {
+
     this.value = this.inputElement.value;
-    emit(this,'change'); // emit change event on root element
+    emit(this, event.type);
```
Hotell, line 63: this is wrong implementation. oninput events are propagated from shadow boundaries. by doing this we are triggering double oninput event

packages/input/index.test.html
```diff
@@ -0,0 +1,69 @@
...
+</head>
+<body>
+
+  <bl-input id="getValue">
+
+  <script>
+
+    suite('<bl-input>', function() {
+
+      var getValue = document.getElementById( 'getValue' );
```
Hotell, line 19: this is very poor and confusing naming

packages/input/index.test.html
```diff
@@ -0,0 +1,69 @@
...
+
+        expect(getValue.value).to.be.equal('c');
+
+      });
+
+      test('emit input', function( done ) {
+
+        var getValueInput = getValue.shadowRoot.querySelector('input');
+
+        getValue.addEventListener( 'input', function() {
```
Hotell, line 41: this test is wrong. you wanna listen on custom element, not on it shadow dom private workings

packages/input/index.test.html
```diff
@@ -0,0 +1,69 @@
...
+
+        getValueInput.dispatchEvent( new Event( 'input' ) );
+
+      });
+
+      test('emit change', function( done ) {
+
+        var getValueInput = getValue.shadowRoot.querySelector('input');
+
+        getValue.addEventListener( 'change', function() {
```
Hotell, line 55: same issue as previous comment

Hotell (changes requested):
My concerns were confirmed by doing this CR. I have to  be part of this development ASAP 🔢

devopsmariocom:
dear @Hotell tests are in progress(bore), if you have time please feel free to contribute to this project, because it's really easy to judge work of others from perspective of god 😉.

PR #114, opened 2017-01-05
Add tagSelector component

### LINKED ISSUE #87: Add Tags

davidrus:
resolves #87

### DIFF
--- a/packages/tag/Tag.scss
+++ b/packages/tag/Tag.scss
@@ -1,4 +1,3 @@
 @import "../generics.shared";
 @import '~blaze/scss/components.buttons';
 @import '~blaze/scss/components.tags';
-@import '~blaze/scss/components.inputs';
--- a/packages/tag/Tag.scss
+++ b/packages/tag/Tag.scss
@@ -1,3 +1,4 @@
 @import "../generics.shared";
 @import '~blaze/scss/components.buttons';
 @import '~blaze/scss/components.tags';
+@import '~blaze/scss/components.inputs';
--- a/packages/tag/Tag.tsx
+++ b/packages/tag/Tag.tsx
@@ -3,7 +3,8 @@ import styles from './Tag.scss';
 
 // public
 interface TagProps extends JSX.HTMLProps<HTMLElement | any> {
-  onTagClose?: Function
+  onTagClose?: Function,
+  label: string,
 }
 
 export class Tag extends Component<TagProps> {
@@ -19,6 +20,15 @@ export class Tag extends Component<TagProps> {
       TAG_CLOSE: 'tagClose'
     }
   }
+  static get props() {
+    return {
+      label: prop.string({
+        attribute: true
+      })
+    }
+  }
+
+  label = '';
 
   private handleClick() {
     emit( this, Tag.events.TAG_CLOSE, {
@@ -29,11 +39,12 @@ export class Tag extends Component<TagProps> {
   }
 
   renderCallback() {
+    const { label } = this;
 
     return [
       <style>{styles}</style>,
       <button type="button" class="c-button c-tag">
-        <slot />
+        { label }
         <span class="c-tag__close" onClick={this.handleClick}>×</span>
       </button>
     ]
--- a/packages/tag/Tag.tsx
+++ b/packages/tag/Tag.tsx
@@ -7,19 +7,25 @@ interface TagProps extends JSX.HTMLProps<HTMLElement | any> {
 }
 
 export class Tag extends Component<TagProps> {
+
+  constructor() {
+    super();
+    this.handleClick = this.handleClick.bind(this);
+  }
+
   static get is() { return 'bl-tag' }
+  static get events() {
+    return {
+      TAG_CLOSE: 'tagClose'
+    }
+  }
 
   private handleClick() {
-    emit( this, 'tagClose', {
+    emit( this, Tag.events.TAG_CLOSE, {
       detail: {
         tag: this
       }
     } )
-  }
-
-  connectedCallback() {
-    super.connectedCallback();
-    this.handleClick = this.handleClick.bind(this);
   }
 
   renderCallback() {
--- a/packages/tag/Tag.tsx
+++ b/packages/tag/Tag.tsx
@@ -1,7 +1,5 @@
-import { h, Component, emit } from 'skatejs';
+import { h, Component, emit, prop } from 'skatejs';
 import styles from './Tag.scss';
-import { Button } from '../button/Button';
-
 
 // public
 interface TagProps extends JSX.HTMLProps<HTMLElement | any> {
@@ -12,7 +10,16 @@ export class Tag extends Component<TagProps> {
   static get is() { return 'bl-tag' }
 
   private handleClick() {
-    emit(this,'tagClose')
+    emit( this, 'tagClose', {
+      detail: {
+        tag: this
+      }
+    } )
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    this.handleClick = this.handleClick.bind(this);
   }
 
   renderCallback() {
--- a/packages/tag/TagSelector.tsx
+++ b/packages/tag/TagSelector.tsx
@@ -1,0 +1,66 @@
+import { h, Component, emit, prop } from 'skatejs';
+import styles from './Tag.scss';
+import { Button } from '../button/Button';
+import { Tag } from './Tag';
+
+
+// public
+interface TagSelectorProps extends JSX.HTMLProps<HTMLElement | any> {
+  onTagChange?: Function,
+  onTagAdd?: Function,
+  delimiter?: string,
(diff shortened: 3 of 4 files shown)


packages/tag/Tag.tsx
```diff
@@ -12,7 +10,16 @@ export class Tag extends Component<TagProps> {
   static get is() { return 'bl-tag' }
 
   private handleClick() {
-    emit(this,'tagClose')
+    emit( this, 'tagClose', {
```
Hotell: Let's not use strings, we all hate them right ? :) 

What I do is this design pattern:
```ts
class MyMama extends Component<Props>{
  static get is(){ return 'yo-mama' }
  static get events(){ return { TAG_CLOSE: 'tagClose' } )
  ....
  private handleClick() {
     emit(this,MamaComponent.events.TAG_CLOSE, payload);
  }
}
```

packages/tag/Tag.tsx
```diff
@@ -12,7 +10,16 @@ export class Tag extends Component<TagProps> {
...
+      detail: {
+        tag: this
+      }
+    } )
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    this.handleClick = this.handleClick.bind(this);
```
Hotell: binding methods to instance should be done in constructor

packages/tag/TagSelector.tsx
```diff
@@ -0,0 +1,88 @@
...
+      delimiter: prop.string({
+        attribute: true
+      }),
+    }
+  }
+
+  tags = [];
+  delimiter = ' '; // default value is space ' '
+
+  private handleInput( event ) {
```
Hotell: event type missing

packages/tag/TagSelector.tsx
```diff
@@ -0,0 +1,88 @@
...
+      detail: {
+        tags: tags
+      }
+    })
+  }
+
+
+  connectedCallback() {
+    super.connectedCallback();
+    this.handleInput = this.handleInput.bind( this );
```
Hotell: binding in ctor

packages/tag/TagSelector.tsx
```diff
@@ -0,0 +1,88 @@
...
+    this.emitNewData( this.tags );
+  }
+
+  private removeTag(event: CustomEvent){
+    const index = this.tags.indexOf( event.detail.tag.innerHTML );
+    this.tags.splice( index, 1 );
+    this.emitNewData( this.tags );
+  }
+
+  private emitNewData( tags ) {
```
Hotell: type of tags missing

packages/tag/TagSelector.tsx
```diff
@@ -0,0 +1,88 @@
...
+    const isDelimiter = lastChar === this.delimiter;
+
+    if ( value && isDelimiter ) {
+      this.addTag(value);
+      event.target.value = '';
+    }
+  }
+
+  private addTag(value:string){
+    this.tags.push(value);
```
Hotell: inner state mutation :(

packages/tag/TagSelector.tsx
```diff
@@ -0,0 +1,88 @@
...
+    });
+
+    return [
+      <style>{styles}</style>,
+      <div class="c-tags">
+        <div class="c-tags__container">
+          {tags}
+        </div>
+        <div class="c-tags__field-container">
+          <input class="c-field"
```
Hotell: we have `bl-input` right? let's use it! :)

packages/tag/Tag.scss
```diff
@@ -1,3 +1,4 @@
 @import "../generics.shared";
 @import '~blaze/scss/components.buttons';
 @import '~blaze/scss/components.tags';
+@import '~blaze/scss/components.inputs';
```
Hotell: we should use existing compoments `bl-input`,`bl-button` so we dont have duplicate styles in various components

packages/tag/TagSelector.tsx
```diff
@@ -0,0 +1,88 @@
...
+        attribute: true
+      }),
+    }
+  }
+
+  tags = [];
+  delimiter = ' '; // default value is space ' '
+
+  private handleInput( event ) {
+    const lastChar = event.target.value.substr(-1);
```
Hotell: we can destructure once instead of accessing long object paths

```ts
const {value} = event.target;
const lastChar = value.substr(-1);
const newValue = value.slice(0, -1).trim();
....
```

packages/tag/TagSelector.tsx
```diff
@@ -0,0 +1,88 @@
...
+      event.target.value = '';
+    }
+  }
+
+  private addTag(value:string){
+    this.tags.push(value);
+    this.emitNewData( this.tags );
+  }
+
+  private removeTag(event: CustomEvent){
```
Hotell: let's be consistent with naming

removeTag --> handleTagClose

because `<Tag onTagClose={}/>`

packages/tag/TagSelector.tsx
```diff
@@ -0,0 +1,88 @@
...
+    const index = this.tags.indexOf( event.detail.tag.innerHTML );
```
Hotell: hmm it isn't safe to rely on DOM, but we have the tag structured via projection, and are emitting whole element instance. So we should maybe redesign the API of that element first. Let's discuss this

packages/tag/TagSelector.tsx
```diff
@@ -0,0 +1,88 @@
...
+    this.tags.splice( index, 1 );
```
Hotell: inner state mutation :(

we should create new array instance and emmit

```ts
const newTags = this.tags.filter((tag,tagIdx)=>tagIdx!== index);
this.emitNewData( newTags );
```

packages/tag/index.demo.tsx
```diff
@@ -1,14 +1,30 @@
...
   private tagClose() {
-    console.log('Tag close emitted', this)
+    console.log( 'Tag close emitted', this )
+    const element = this;
+    element.parentNode.removeChild( element );
+  }
+
+  private tagChange( event ) {
+    console.log( 'Tag change:', event.detail.tags );
+    this.tags = [].concat( event.detail.tags );
```
Hotell: you can do shorthand via spread ( shallow copy )
`this.tags = [ ...event.detail.tags ];`

Hotell (changes requested):

packages/tag/TagSelector.tsx
davidrus: We will talk about this today...

packages/tag/TagSelector.tsx
davidrus: Changed, but we must fix bl-input to provide onInput properly

davidrus:
Ready to re-review

packages/tag/TagSelector.tsx
```diff
@@ -58,7 +59,8 @@ export class TagSelector extends Component<TagSelectorProps> {
   }
 
   private handleTagClose(event: CustomEvent) {
-    const index = this.tags.indexOf( event.detail.tag.innerHTML );
+    const target = event.target as Tag;
+    const index = this.tags.indexOf( target.label );
```
Hotell: now we can do cleaner removal:

instead of:
```ts
const index = this.tags.indexOf( target.label );
const newTags = this.tags.filter( (tag, tagIdx) => tagIdx !== index );
```
this:
```ts
const newTags = this.tags.filter( (tag) => tag !== target.label );
```

packages/tag/TagSelector.tsx
```diff
@@ -40,15 +40,16 @@ export class TagSelector extends Component<TagSelectorProps> {
   delimiter = ' '; // default value is space ' '
```
Hotell, line 40: pls add explicit type annotation to tag. In our case `tags: string[] = []`

packages/tag/index.demo.tsx
```diff
@@ -31,10 +31,10 @@ export class Demo extends Component<void> {
         <legend>{Tag.is}</legend>
 
         <div>
-          <Tag onTagClose={this.tagClose}>Tag one</Tag>
-          <Tag onTagClose={this.tagClose}>Tag two</Tag>
-          <Tag onTagClose={this.tagClose}>Tag three</Tag>
-          <Tag onTagClose={this.tagClose}>Tag four</Tag>
+          <Tag onTagClose={this.tagClose} label="Taone"></Tag>
```
Hotell: don't use closing tags if element has no children. this rule should be done in `tslint` ( currently tslint definition is very poor. I'll update it later )

instead:
`<Tag onTagClose={this.tagClose} label="Taone"></Tag>`

do:
`<Tag onTagClose={this.tagClose} label="Taone" />`

Hotell (changes requested):
good job, just few minor tweaks

davidrus:
done

Hotell (approved):

PR #119, opened 2017-01-09
docs(README): update readme with initial setup for development

Hotell:


davidrus (approved):
For the first initial readme is enaugh. Good job 👍

PR #121, opened 2017-01-09
add <bl-calendar> component

### LINKED ISSUE #16: implement <bl-calendar>
http://blazecss.com/components/calendars/

This may later be used in Date Picker & Date Time Picker.

However **Calendar** component should be plain calendar so that it can be used independently.

davidrus:
Fixes #16

### DIFF
--- a/packages/_helpers/buildFormatLocale.ts
+++ b/packages/_helpers/buildFormatLocale.ts
@@ -7,6 +7,7 @@ export type LocaleType = {
   meridiemUppercase?: string[],
   meridiemLowercase?: string[],
   meridiemFull?: string[],
+  todayButtonText?: string,
 };
 
 const defaultLocale: LocaleType = {
--- a/packages/_helpers/buildFormatLocale.ts
+++ b/packages/_helpers/buildFormatLocale.ts
@@ -20,16 +20,13 @@ const defaultLocale: LocaleType = {
   meridiemFull: ['a.m.', 'p.m.'],
 };
 
-export function buildFormatLocale ( locale?: LocaleType ) {
+export function buildFormatLocale ( customLocale?: LocaleType ) {
   // Note: in English, the names of days of the week and months are capitalized.
   // If you are making a new locale based on this one, check if the same is true for the language you're working on.
   // Generally, formatted dates should look like they are in the middle of a sentence,
   // e.g. in Spanish language the weekdays and months should be in the lowercase.
 
-  let mergedLocale: LocaleType = defaultLocale;
-  for ( let key in locale ) {
-    mergedLocale[ key ] = locale[ key ];
-  }
+  let mergedLocale: LocaleType = Object.assign( {}, defaultLocale, customLocale );
 
   const formatters = {
     // Month: Jan, Feb, ..., Dec
@@ -65,9 +62,7 @@ export function buildFormatLocale ( locale?: LocaleType ) {
     };
   });
 
-  return {
-    formatters: formatters
-  };
+  return { formatters };
 }
 
 function ordinal (num: number) {
--- a/packages/_helpers/buildFormatLocale.ts
+++ b/packages/_helpers/buildFormatLocale.ts
@@ -1,0 +1,86 @@
+export type LocaleType = {
+  months3char?: string[],
+  monthsFull?: string[],
+  weekdays2char?: string[],
+  weekdays3char?: string[],
+  weekdaysFull?: string[],
+  meridiemUppercase?: string[],
+  meridiemLowercase?: string[],
+  meridiemFull?: string[],
+};
+
+const defaultLocale: LocaleType = {
+  months3char: ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'],
+  monthsFull: ['January', 'February', 'March', 'April', 'May', 'June', 'July', 'August', 'September', 'October', 'November', 'December'],
+  weekdays2char: ['Su', 'Mo', 'Tu', 'We', 'Th', 'Fr', 'Sa'],
+  weekdays3char: ['Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat'],
+  weekdaysFull: ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'],
+  meridiemUppercase: ['AM', 'PM'],
+  meridiemLowercase: ['am', 'pm'],
+  meridiemFull: ['a.m.', 'p.m.'],
+};
+
+export function buildFormatLocale ( locale?: LocaleType ) {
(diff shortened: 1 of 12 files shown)


packages/_helpers/buildFormatLocale.ts
```diff
@@ -0,0 +1,86 @@
...
+  meridiemFull: ['a.m.', 'p.m.'],
+};
+
+export function buildFormatLocale ( locale?: LocaleType ) {
+  // Note: in English, the names of days of the week and months are capitalized.
+  // If you are making a new locale based on this one, check if the same is true for the language you're working on.
+  // Generally, formatted dates should look like they are in the middle of a sentence,
+  // e.g. in Spanish language the weekdays and months should be in the lowercase.
+
+  let mergedLocale: LocaleType = defaultLocale;
```
Hotell: `const mergedLocale = Object.assign({},defaultLocale,locale)`

packages/_helpers/buildFormatLocale.ts
```diff
@@ -0,0 +1,86 @@
...
+  // Generate ordinal version of formatters: M -> Mo, D -> Do, etc.
+  const ordinalFormatters = ['M', 'D', 'DDD', 'd', 'Q', 'W'];
+  ordinalFormatters.forEach(function (formatterToken) {
+    formatters[formatterToken + 'o'] = function (date, formatters) {
+      return ordinal(formatters[formatterToken](date));
+    };
+  });
+
+  return {
+    formatters: formatters
```
Hotell: `return {formatters}`

davidrus:
PR is ready to review, thanks guys!! 👍

packages/calendar/Calendar.tsx
```diff
@@ -0,0 +1,202 @@
...
+import { buildFormatLocale, LocaleType } from '../_helpers/buildFormatLocale';
+
+interface CalendarProps extends JSX.HTMLProps<HTMLElement | any> {
+  selectedDate?: string,
+  onDateChange?: Function,
+  i18n?: {
+    monthsFull: string[],
+    weekdays2char: string[],
+  },
+  weekStartsOn?: number,
```
hopkins-tk: Enumeration: MONDAY, SUNDAY, ... would be better as I would have no idea as a user, what 0/1/... mean :)

packages/calendar/Calendar.tsx
```diff
@@ -0,0 +1,211 @@
...
+import * as format from 'date-fns/format';
+import * as isToday from 'date-fns/is_today';
+import * as parse from 'date-fns/parse';
+import * as isEqual from 'date-fns/is_equal';
+import { css } from '../_helpers/css';
+import { Button } from '../button/Button';
+import { buildFormatLocale, LocaleType } from '../_helpers/buildFormatLocale';
+
+interface CalendarProps extends JSX.HTMLProps<HTMLElement | any> {
+  selectedDate?: string,
```
hopkins-tk: Could not the property be _Date_?
- @Hotell  what do you think?

webpack.config.js
```diff
@@ -54,7 +54,7 @@ module.exports = ( env ) => {
      * See: https://webpack.js.org/configuration/devtool/#devtool
      * See: https://github.com/webpack/docs/wiki/build-performance#sourcemaps
      */
-    devtool: ifProd( 'source-map', 'cheap-module-source-map' ),
+    devtool: 'source-map',
```
devopsmariocom: any reason for this change?

packages/calendar/Calendar.tsx
```diff
@@ -0,0 +1,211 @@
...
+  onDateChange?: Function,
```
hopkins-tk: I would like to have type definition for that.

What about something like:
```javascript
interface DateChangedEventDetail {
  readonly date: Date;
}
export interface DateChangedEvent extends Event {
  readonly detail: DateChangedEventDetail
}
```
and `onDateChange?: DateChangedEvent`

packages/calendar/Calendar.tsx
```diff
@@ -0,0 +1,211 @@
...
+  i18n?: {
+    monthsFull: string[],
+    weekdays2char: string[],
```
hopkins-tk, line 27: Those two are string arrays - in case of months there should not be any issue, but in case of weekdays - the order might not be clear as some countries start with Monday, some with Sunday and therefore the people might be inclined to add list in their national order.

If there was an object structure being used, it would be absolutely clear, what one should put there.

webpack.config.js
davidrus: yes, but actually this shouldn't be part of this PR, I will create new PR with issue

packages/calendar/Calendar.tsx
```diff
@@ -0,0 +1,211 @@
...
+    this.prevMonth = this.prevMonth.bind( this );
+    this.nextMonth = this.nextMonth.bind( this );
+    this.setDate = this.setDate.bind( this );
+  }
+
+  static get is() { return 'bl-calendar'; }
+
+  static get props() {
+    return {
+      days: prop.array(),
```
hopkins-tk, line 49: What is a purpose of `days` property? - do we want to have it configurable by users?

webpack.config.js
davidrus: #124

packages/calendar/Calendar.tsx
```diff
@@ -0,0 +1,211 @@
...
+        <button class="c-calendar__control" onClick={this.nextYear}>›</button>
+
+        <button class="c-calendar__control" onClick={this.prevMonth}>‹</button>
+        <div class="c-calendar__header">{this.format( new Date( year, month ), 'MMMM' )}</div>
+        <button class="c-calendar__control" onClick={this.nextMonth}>›</button>
+
+        {weekDays}
+        {days}
+
+        <Button onClick={this.setDateHandler()} block>Today</Button>
```
hopkins-tk: Localization is missing here.

hopkins-tk (changes requested):

packages/calendar/Calendar.tsx
davidrus: I dont know when we want to do it. Because `onDateChanged` in interface is for user which want to use component like `<Calendar onDateChange={this.dateChangeHandler}  />` where  **dateChangeHandler** is their function, not event.

packages/calendar/Calendar.tsx
davidrus: We need to call renderCallback() when days or (years|months) are changed, so this is reason what is here.

packages/calendar/Calendar.tsx
davidrus: I forgot that 🇨🇿 🇺🇸 🇩🇪 🇬🇧  = 🏴

packages/calendar/Calendar.tsx
davidrus: We can talk about this at stand-up :)

davidrus:
PR is ready to re-review. @hopkins-tk

hopkins-tk (approved):

PR #122, opened 2017-01-09
feat(button): add support for block display

### LINKED ISSUE #120: bl-button - new css class
Add support to display button with css `c-button--block` to show it with full width in container. We need this to show button in [calendar](http://blazecss.com/components/calendars/) for example.

davidrus:
resolve #120

### DIFF
--- a/packages/button/package.json
+++ b/packages/button/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bl-button",
-  "version": "1.1.0",
+  "version": "1.1.1",
   "main": "dist/main.js",
   "license": "MIT",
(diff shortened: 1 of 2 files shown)


packages/button/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "bl-button",
-  "version": "1.1.0",
+  "version": "1.1.1",
```
Hotell: 1.2.0 pls. This is a feat

Hotell (changes requested):

Hotell (approved):

PR #130, opened 2017-01-19
refactor(packages/*): update to latest skate and fix TS definitions

Hotell:
This PR does:
- adds latest skate with new TS definitions
- removes our custom skate definitions
- fixes all TS errors 
- refactors Alert component to make it as a Style guide example

prerequisite to merge, is this PR: https://github.com/skatejs/skatejs/pull/1022


packages/alert/Alert.tsx
```diff
@@ -46,13 +58,17 @@ export class Alert extends Component<AlertProps> {
 
     return [
       <style>{styles}</style>,
-      isOpen && <div className={className}>
-        <Button close onClick={this.close}>×</Button>
-        <slot />
-      </div>
+      isOpen && (
```
davidrus, line 61: Why do you add brackets (just asking for me :) )

Hotell: Right, this pattern will be enforced by TS lint.
Its just how JS works, if I would omit the brace, JS compiler will on the line `isOpen &&` ( in this case it would trigger an error ) end. by using brace we are telling it that this is not the end of execution, and thanks to that it allow us to format JSX on  separate line which makes the code more readable )

hopkins-tk (approved):

packages/alert/Alert.tsx
davidrus: Thank You!!!

davidrus (approved):

Hotell:
can you review pls @elmariofredo ASAP? we need this to be merged to continue with clean slate

packages/alert/Alert.tsx
Hotell: you're welcome! 🍺

Hotell:
@elmariofredo can you review pls? we are waiting just for your approval. I you don't have time or don't care , next please at least write something down :) :D

devopsmariocom (approved):

PR #135, opened 2017-01-23
chore(badge): small refactor && tests

davidrus:
Mainly I've added tests, please take a look.

### DIFF
--- a/packages/badge/Badge.test.tsx
+++ b/packages/badge/Badge.test.tsx
@@ -10,73 +10,85 @@ describe( Badge.is, () => {
 
       expect( customElements.get( Badge.is ) ).toBe( Badge );
 
-    } );
+    });
 
     it( `should render via JSX IntrinsicElement`, () => {
 
       return mount(
-        <bl-badge/>
-      ).wait( element => {
+        <bl-badge />
+      ).wait(( element ) => {
 
         expect( element.node.localName ).toBe( Badge.is );
 
-      } );
+      });
 
-    } );
+    });
 
     it( `should render via JSX class`, () => {
 
       return mount(
-        <Badge/>
-      ).wait( ( element ) => {
+        <Badge />
+      ).wait(( element ) => {
 
         expect( element.has( '.c-badge' ) ).toBe( true );
 
-      } );
+      });
 
-    } );
-  } );
+    });
+  });
 
   describe( 'API', () => {
 
-    it( 'should render with correct color', () => {
+    describe( `[color]`, () => {
 
-      return mount(
-        <bl-badge attributes={{color:'warning'} as any}></bl-badge>
-      ).wait( ( element ) => {
+      it( 'should render with correct color', () => {
 
-        expect( element.has( '.c-badge--warning' ) ).toBe( true );
+        return mount(
+          <bl-badge color="warning" />
+        ).wait(( element ) => {
 
-      } );
+          expect( element.has( '.c-badge--warning' ) ).toBe( true );
 
-    } );
+        });
 
-    it( 'should render ghosted', () => {
+      });
 
-      return mount(
-        <bl-badge ghost></bl-badge>
-      ).wait( ( element ) => {
+    });
 
-        expect( element.has( '.c-badge--ghost' ) ).toBe( true );
+    describe( `[ghost]`, () => {
 
-      } );
+      it( 'should render ghosted', () => {
 
-    } );
+        return mount(
+          <bl-badge ghost />
+        ).wait(( element ) => {
 
-    it( 'should render rounded', () => {
+          expect( element.has( '.c-badge--ghost' ) ).toBe( true );
 
-      return mount(
-        <bl-badge rounded></bl-badge>
-      ).wait( ( element ) => {
+        });
 
-        expect( element.has( '.c-badge--rounded' ) ).toBe( true );
+      });
 
-      } );
+    });
 
-    } );
+    describe( `[rounded]`, () => {
 
-  } );
+      it( 'should render rounded', () => {
 
-} );
+        return mount(
+          <bl-badge rounded />
+        ).wait(( element ) => {
+
+          expect( element.has( '.c-badge--rounded' ) ).toBe( true );
+
+        });
+
+      });
+
+    });
+
+  });
+
+});
 
 
--- a/packages/badge/Badge.test.tsx
+++ b/packages/badge/Badge.test.tsx
@@ -90,5 +90,3 @@ describe( Badge.is, () => {
   });
 
 });
-
-
--- a/packages/badge/Badge.test.tsx
+++ b/packages/badge/Badge.test.tsx
@@ -1,7 +1,7 @@
-import { Badge } from './index';
+import * as expect from 'expect';
 import { mount } from 'bore';
 import { h } from '../_helpers';
-import * as expect from 'expect';
+import { Badge } from './index';
 
 describe( Badge.is, () => {
 
@@ -37,11 +37,11 @@ describe( Badge.is, () => {
     });
   });
 
-  describe( 'API', () => {
+  describe( `API`, () => {
 
     describe( `[color]`, () => {
 
-      it( 'should render with correct color', () => {
+      it( `should render with correct color`, () => {
 
         return mount(
           <bl-badge color="warning" />
@@ -57,7 +57,7 @@ describe( Badge.is, () => {
 
     describe( `[ghost]`, () => {
 
-      it( 'should render ghosted', () => {
+      it( `should render ghosted`, () => {
 
         return mount(
           <bl-badge ghost />
@@ -73,7 +73,7 @@ describe( Badge.is, () => {
 
     describe( `[rounded]`, () => {
 
-      it( 'should render rounded', () => {
+      it( `should render rounded`, () => {
 
         return mount(
           <bl-badge rounded />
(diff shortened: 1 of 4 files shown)


packages/badge/Badge.test.tsx
```diff
@@ -7,12 +7,52 @@ describe( Badge.is, () => {
...
+        <Badge color="brand">Brand</Badge>
+        <Badge color="info">Info</Badge>
+        <Badge color="warning">Warning</Badge>
+        <Badge color="success">Success</Badge>
+        <Badge color="error">Error</Badge>
+      </div>
+    ).waitFor((wrapper) => {
+      const badges = wrapper.all(Badge);
+      return (
+        !!badges[0].node.shadowRoot
```
devopsmariocom: I'm not sure if we need to test all colors a we have helper for that, also testing just first shadowRoot doesn't feels right. I would either return true or check all.

devopsmariocom (changes requested):

packages/badge/Badge.test.tsx
davidrus: It works with `return:true`. I replaced this with it.

packages/badge/Badge.test.tsx
```diff
@@ -1,18 +1,51 @@
...
-    console.warn('Missing tests for Badge!');
-
     return mount(
       <Badge/>
-    ).wait();
+    ).wait((element) => element.has('c-badge') );
 
   });
 
+  it( 'should render rounded', () => {
```
Hotell: we need also test intrinsic element.

packages/badge/Badge.tsx
```diff
@@ -1,14 +1,22 @@
...
   color?: ColorType,
   rounded?: boolean,
   ghost?: boolean,
 }
 
+// extend JSX.IntrinsicElements namespace with our definition
+declare global {
+  namespace JSX {
+    interface IntrinsicElements {
+      'bl-badge': BadgeProps,
```
Hotell: and `& HTMLSpanElement`

packages/badge/index.demo.tsx
```diff
@@ -6,7 +6,7 @@ export class Demo extends Component<void> {
 
   renderCallback() {
     return [
-      <style></style>,
+      <style/>,
```
Hotell: if we don't use style or particular tag, just remove it

packages/badge/Badge.test.tsx
Hotell: hmm I don't see any assertions in whole test :D

so this whole suit does nothing basically

Hotell (changes requested):

packages/badge/Badge.test.tsx
davidrus: What do you mean 'nothing'? 😄  ... This test is only for css class with this bool flag. Nothing more...

davidrus:
Changes done

packages/badge/Badge.test.tsx
devopsmariocom: Captain here, this pattern is taken from bore readme. Even that martin is right and test should usually have at least one assertion in this case it is valid way of testing asynchronous code. I don't see any advatages of adding is.equal(true) everywhere.

devopsmariocom (approved):

packages/badge/Badge.test.tsx
Hotell: it has nothing to do with testing asynchronous code. That's implementation detail of bore that `wait` returns Promise and that mocha is able to consume it.  that `mount` call is just "another" `before` hook specific to bore et all.

I want to have there assertions, to be consistent with all testing, although sometimes it will be on line more.

```
 it( 'should render intrinsic element', () => {
   return mount(
     <bl-badge></bl-badge>
   ).wait((element) => {
       const actual = element.has('c-badge'); 
       const expected = true;
       expect( actual ).to.equal( expected );
    });
  
});
```

Hotell (changes requested):

packages/badge/Badge.test.tsx
```diff
@@ -47,11 +59,17 @@ describe( Badge.is, () => {
       </div>
     ).waitFor(() => true ).then((wrapper) => {
       const badges = wrapper.all(Badge);
-      return  badges[0].has('c-badge__brand') &&
-              badges[1].has('c-badge__info') &&
-              badges[2].has('c-badge__warning') &&
-              badges[3].has('c-badge__success') &&
-              badges[4].has('c-badge__error');
+      const actualBrand = badges[0].node.shadowRoot.querySelectorAll('.c-badge--brand').length;
```
Hotell: mmm before my next  comment try using `one`, it should walk inside shadowRoot

maybe we should create a private function to access shadowRoot instead of repeating zilion times

smthng like `getShadowRoot(element){}`

Hotell (changes requested):

packages/badge/Badge.test.tsx
davidrus: Take a look to combination 'all' which return wrappers and 'one'.

packages/badge/Badge.test.tsx
```diff
@@ -1,18 +1,80 @@
-import { Badge } from './Badge';
-
+import { Badge } from './index';
 import { h, mount } from 'bore';
-import { emit } from 'skatejs';
 
-describe( Badge.is, () => {
+describe.only( Badge.is, () => {
```
Hotell: ehm ehm :D 🗡

davidrus: I kill you :D :D :D

davidrus: https://www.youtube.com/watch?v=GBvfiCdk-jc

packages/badge/Badge.test.tsx
```diff
@@ -1,18 +1,76 @@
...
+    return mount(
+      <div>
+        <Badge color="brand">Brand</Badge>
+        <Badge color="info">Info</Badge>
+        <Badge color="warning">Warning</Badge>
+        <Badge color="success">Success</Badge>
+        <Badge color="error">Error</Badge>
+      </div>
+    ).waitFor((wrapper: any) => {
+      const badges = wrapper.all(Badge);
```
Hotell: hmm maybe we should create helper for this, Smthng like:

```ts
function allElementsShadowReady(...elements){
     return elements.every( elem => elem.node.shadowRoot )
}
```

Usage
```ts
.waitFor((wrapper) => {
  return allShalowElementsShadowReady(wrapper.all(Badge))
})
.then()
```

cc @elmariofredo  @hopkins-tk

packages/badge/Badge.test.tsx
```diff
@@ -1,18 +1,76 @@
...
 
   it( 'should render', () => {
 
-    console.warn('Missing tests for Badge!');
-
     return mount(
       <Badge/>
-    ).wait();
+    ).wait((wrapper) => {
+      expect(wrapper.has('.c-badge')).to.equal(true);
```
devopsmariocom: sorry for one additional but let's use chai bdd style guide here http://chaijs.com/api/bdd/#method_true which means `.to.be.true`

devopsmariocom (changes requested):

Hotell (changes requested):

packages/badge/Badge.test.tsx
davidrus: What do you say guys? @elemementor @hopkins-tk && @elmariofredo

hopkins-tk: So we have agreed that `allElementsShadowReady` whould be helpful and should definitely be created, but this test should probably cover just single color.

And there should be separate test(s) for the helper itself.

Thus avoiding need for `allElementsShadowReady` in this test.

hopkins-tk (changes requested):

packages/badge/Badge.test.tsx
davidrus: affirmative sir!! done....

packages/badge/Badge.test.tsx
```diff
@@ -1,18 +1,66 @@
-import { Badge } from './Badge';
-
+import { Badge } from './index';
 import { h, mount } from 'bore';
-import { emit } from 'skatejs';
 
-describe( Badge.is, () => {
+describe( `${Badge.is}`, () => {
```
hopkins-tk: Why replacing `Badge.is` with `${Badge.is}` - isn't it the same?

hopkins-tk (changes requested):

davidrus:
Rebased and tests are refactored to use `expect` and our `h`.

devopsmariocom (approved):

Hotell (approved):

packages/badge/Badge.test.tsx
```diff
@@ -1,17 +1,94 @@
...
+
+        });
+
+      });
+
+    });
+
+    describe( `[rounded]`, () => {
+
+      it( 'should render rounded', () => {
```
hopkins-tk: Please unify to _string literal_ everywhere (describe & it).

hopkins-tk (changes requested):

packages/badge/Badge.test.tsx
davidrus: Nooooooo!!!!

hopkins-tk (approved):

devopsmariocom:
Tests are passing ;)

PR #137, opened 2017-01-24
Components packaging example + minor refactor

Hotell:


packages/calendar/package.json
```diff
@@ -11,5 +11,8 @@
     "build": "../../node_modules/.bin/rimraf dist && ../../node_modules/.bin/webpack --config ../../webpack.config.js --env.prod --env.element=calendar",
     "build:test": "../../node_modules/.bin/rimraf dist && ../../node_modules/.bin/webpack --config ../../webpack.config.js --env.test --env.element=calendar",
     "build:test:watch": "../../node_modules/.bin/webpack-dev-server --config ../../webpack.config.js --env.test --env.element=calendar"
+  },
+  "dependecies":{
+    "bl-button": "1.2.0"
```
devopsmariocom, line 16: does this work with `lerna bootstrap` ?

Hotell: I don't know :) it should work

Hotell: I've added stuff how should reference package look like. If it would makes problem for current build we can remove it ofc

devopsmariocom (approved):

davidrus (approved):

hopkins-tk (approved):

PR #138, opened 2017-01-24
chore(alert): use explicit import

devopsmariocom:
as './' is interpreted as normal package and thus it use 'packages/alert/dist/main.js' as defined in package.json


devopsmariocom:
other wise it will fail on missing __extense when dist directory is present for alert package

Hotell (approved):

davidrus (approved):

PR #139, opened 2017-01-25
refactor(tooltip): refactoring to the new structure + tests

hopkins-tk:
- tooltip refactored to the new style
- tests added

### DIFF
--- a/packages/tooltip/Tooltip.test.tsx
+++ b/packages/tooltip/Tooltip.test.tsx
@@ -1,4 +1,6 @@
-import { h, mount } from 'bore';
+import { h } from '../_helpers';
+import { mount } from 'bore';
+import * as expect from 'expect';
 import { Tooltip } from './index';
 
 describe( Tooltip.is, () => {
@@ -6,7 +8,7 @@ describe( Tooltip.is, () => {
   describe( `Custom element`, () => {
     it( `should be registered`, () => {
 
-      expect( customElements.get( Tooltip.is ) ).to.equal( Tooltip );
+      expect( customElements.get( Tooltip.is ) ).toBe( Tooltip );
 
     });
 
@@ -14,9 +16,9 @@ describe( Tooltip.is, () => {
 
       return mount(
         <bl-tooltip label="hello"/>
-      ).wait( element => {
+      ).wait( ( element ) => {
 
-        expect( element.node.localName ).to.equal( Tooltip.is );
+        expect( element.node.localName ).toBe( Tooltip.is );
 
       });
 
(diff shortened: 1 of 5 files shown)


packages/tooltip/Tooltip.test.tsx
```diff
@@ -1,17 +1,49 @@
...
 
   it( 'should render', () => {
 
-    console.warn('Missing tests for Tooltip!');
-
     return mount(
       <Tooltip label="hello"/>
-    ).wait();
+    ).wait((element) => {
+      expect( element.has('.c-tooltip') ).to.be.true;
```
Hotell: we use just `to.equal` or `to.deep.equal` for assertions

packages/tooltip/Tooltip.tsx
```diff
@@ -9,34 +9,43 @@ const TooltipTypes = {
...
+type TooltipProps = {
+  type?: TooltipTypesType,
   label: string,
-}
+};
+
+declare global {
+  namespace JSX {
+    interface IntrinsicElements {
+      'bl-tooltip': TooltipProps & Partial<HTMLElement>,
```
Hotell, line 21: root node is `span` so `HTMLSpanElement`

packages/tooltip/Tooltip.test.tsx
```diff
@@ -1,17 +1,49 @@
...
+      <bl-tooltip label="hello"></bl-tooltip>
+    ).wait((element) => {
+      expect( element.has('.c-tooltip') ).to.be.true;
+      expect( element.has('.c-tooltip--right') ).to.be.true;
+      expect( element.one('span').node.getAttribute('aria-label') ).to.equals( 'hello' );
+    });
+
+  });
+
+  it( 'should render with top tooltip', () => {
```
Hotell: this and following test are doing the same assertions. Unit test should be clean as our codebase, so extract assertion to one function helper.

also the description of `it` isn't really telling anything.

We should probably structure every test to 2 sub `describe` one for `JSX` and one for "raw" `html` custom element usage

Hotell (changes requested):

packages/tooltip/Tooltip.test.tsx
hopkins-tk: Good idea with the two `describe` - I was thinking how we will want to structure this.

packages/tooltip/Tooltip.test.tsx
hopkins-tk: Great :) we have been discussing this yesterday with @elmariofredo  & @davidrus so we can take the opportunity and agree what should be the preferred version.

packages/tooltip/Tooltip.tsx
hopkins-tk: So any other component, that has DIV as a root element should have `HTMLDivElement`?
- is this really correct?
- if we don't pass any attributes to the root element

Hotell: ok it was late and I was tired :D so nevermind. Let's stick with `HTMLElement` for all intrinsic elements

packages/tooltip/Tooltip.test.tsx
```diff
@@ -1,18 +1,62 @@
-import { Tooltip } from './Tooltip';
-
+import { Tooltip } from './index';
 import { h, mount } from 'bore';
-import { emit } from 'skatejs';
+import {WrappedNode} from "bore";
```
Hotell: pls merge with `bore` imports Webstorm has it and VSCode as well

packages/tooltip/Tooltip.test.tsx
```diff
@@ -1,18 +1,62 @@
...
+      return mount(
+        <bl-tooltip label="hello"></bl-tooltip>
+      ).wait(expectDefaultRender);
+
+    });
+
+    it( 'should render with top tooltip', () => {
+
+      return mount(
+        <bl-tooltip label="hello" type="top"></bl-tooltip>
```
Hotell: just a reminder how JSX works ( with bore `h` ):
` <Tooltip label="hello" type="top" />` transforms to:
- `h( Tooltip, {label:'hello', type:'top'} )` 
- which is created by `new Tooltip()`

`<bl-tooltip label="hello" type="top"></bl-tooltip>` transforms to:
- `h('bl-tooltip',{label:'hello', type:'top'})`
- which is created by `document.createElement`

Hotell (changes requested):

packages/tooltip/Tooltip.tsx
```diff
@@ -9,34 +9,47 @@ const TooltipTypes = {
   bottom: 'bottom',
 };
 
-type TooltipTypesType = typeof TooltipTypes;
+type TooltipTypesType = keyof typeof TooltipTypes;
```
Hotell: we dont need that suffix TooltipTypes`Type` in name. We now that it is a type alias already ;)

Hotell (changes requested):
just minor tweaks. Otherwise LGMT. and lets wait for new bore release. Good job @hopkins-tk !

hopkins-tk:
This is failing because of skatejs/bore#11

Hotell:
We should temporarely monkey patch bore so we don't get blocked by it

hopkins-tk:
Rebased & refactored to the newest version of tests we have agreed on.

packages/tooltip/Tooltip.test.tsx
```diff
@@ -1,16 +1,99 @@
...
+          <bl-tooltip label="hello <&> Hi" />
+        ).wait(( element ) => {
+
+          expect( element.one('span').node.getAttribute('aria-label') ).toBe( 'hello <&> Hi' );
+
+        });
+
+      });
+
+      it( `should render with correct label using property`, () => {
```
Hotell: this test is doing exactly the same as last one

Hotell: probably you wanted to us `attributes` in the first one

packages/tooltip/Tooltip.tsx
```diff
@@ -9,34 +9,45 @@ const TooltipTypes = {
...
-type TooltipTypesType = typeof TooltipTypes;
-
 // public
-interface TooltipProps {
-  type?: keyof TooltipTypesType,
-  label: string,
-}
+type TooltipProps = {
+  type?: keyof typeof TooltipTypes,
+  label?: string,
```
Hotell: label should be required

Hotell (changes requested):

Hotell (approved):

PR #151, opened 2017-02-01
refactor(input): remove inner state of input & new onchange propagation

### LINKED ISSUE #115: bl-input api issues
few things that we found during PR of #114 

- `bl-input` needs to support only `change` event so we have consistent API
-  so `onInput` needs to be stoped ( so user cannot use "unsupported" onInput event api )
- inputElement ref, needs to be `private` so user cannot access it and call it a day :D

davidrus:
affects: bl-input

resolves #115

### DIFF
--- a/packages/input/Input.test.tsx
+++ b/packages/input/Input.test.tsx
@@ -1,7 +1,6 @@
 import { mount, h } from 'bore';
 import * as expect from 'expect';
 import { Input } from './index';
-import { emit } from 'skatejs';
 
 describe( Input.is, () => {
 
@@ -159,12 +158,12 @@ describe( Input.is, () => {
           changeTriggered = true;
         };
 
-        return mount( <bl-input value="water" events={{ huh: handleChange, change: handleChange } as any} /> )
+        return mount( <bl-input value="water" events={{ change: handleChange }} /> )
           .wait(( element ) => {
 
             const input = element.one( 'input' ).node as HTMLInputElement;
             input.value = 'splash';
-            emit( input, 'input' );
+            input.dispatchEvent( new Event( 'input' ) );
             expect( changeTriggered ).toBe( true );
 
           });
--- a/packages/input/Input.test.tsx
+++ b/packages/input/Input.test.tsx
@@ -1,11 +1,12 @@
-import { h } from '../_helpers';
-import { mount } from 'bore';
+import { mount, h } from 'bore';
 import * as expect from 'expect';
 import { Input } from './index';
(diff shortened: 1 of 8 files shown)


davidrus:
Guys, please take a look and we should discuss about testing of event propagation.

packages/input/Input.tsx
```diff
@@ -29,38 +28,60 @@ type Props = {
...
       change: 'change',
     };
   }
 
   valid: string;
-  value = '';
+  value: string;
   inputSize: Size;
   placeholder: string;
   type = 'text';
```
hopkins-tk, line 96: Aren't you facing issues with the default value & polyfills?

packages/input/Input.tsx
```diff
@@ -91,30 +110,21 @@ export class Input extends Component<InputProps> {
...
         disabled={disabled}
       />
     ];
   }
 
   private propagateOnChange( event: Event ) {
-    this.setValue();
-    emit( this, Input.events.change ); // emit change event on root element
+    // emit change event on root element
+    const input: Partial<HTMLInputElement> = event.target;
```
Hotell: we need to stop propagation of native event, otherwise our api would be leaky - this implementation will also trigger `onInput` regular DOM listener.


```ts
e.stopImmediatePropagation():
emit( this, Input.events.change, { detail: { data: input.value } });
```

packages/input/Input.test.tsx
```diff
@@ -1,48 +1,155 @@
...
+
+        expect( element.has( '.c-field' ) ).toBe( true );
+
+      });
+
+    });
   });
 
-  it( 'should set value and emit on input', ( done ) => {
+  describe( `API`, () => {
```
Hotell: there is no test for emitting `onChange`. Pls add, don't hesitate to ask me, if you don't know how. thx

Hotell (changes requested):
overall good job! just few nit picks

👍

packages/input/Input.test.tsx
davidrus: Yes, when you have one minute of your time, please come here and we will create spec for it.

packages/input/Input.tsx
davidrus: no :)

Hotell (approved):

Hotell:
sorry my bad, we have to switch back to `emit` instead of manual `dispatchEvent` because of http://stackoverflow.com/questions/27643181/ie-11-dispatchevent

PR #153, opened 2017-02-03
chore(webpack): improvements of build artifacts

hopkins-tk:
Build of the whole project will now produce:
- **icons-...** asset
- **index.html**
- **index.min.js**
- **index-with-dependencies.min.js**
- **styles.min.js**


index.test.html
```diff
@@ -2,12 +2,12 @@
 <html>
 <head>
   <meta charset="utf-8">
-  <script src="./dist/test-helpers.js"></script>
+  <script src="./dist/test-helpers.min.js"></script>
```
devopsmariocom: why do we use minimalised versions in tests? any reason for that?

packages/accordion/index.with.dependencies.ts
```diff
@@ -0,0 +1,2 @@
+import '../../vendors';
```
devopsmariocom: can we somehow set single entrypoint for multiple scripts? I'm sure that this was passible in webpack 1 see https://webpack.github.io/docs/multiple-entry-points.html

devopsmariocom (changes requested):

packages/tooltip/package.json
```diff
@@ -10,6 +10,6 @@
     "test:watch:firefox": "../../node_modules/.bin/concurrently \"npm run build:test:watch\" \"WCT_INDEX=./index.watch.test.html ../../node_modules/.bin/wct -p -l firefox\"",
     "build": "../../node_modules/.bin/rimraf dist && ../../node_modules/.bin/webpack --config ../../webpack.config.js --env.prod --env.element=tooltip",
     "build:test": "../../node_modules/.bin/rimraf dist && ../../node_modules/.bin/webpack --config ../../webpack.config.js --env.test --env.element=tooltip",
-    "build:test:watch": "../../node_modules/.bin/webpack-dev-server --config ../../webpack.config.js --env.test --env.element=tooltip"
+    "build:test:watch": "../../node_modules/.bin/webpack-dev-server --config ../../webpack.config.js --env.test --host 0.0.0.0 --env.element=tooltip"
```
devopsmariocom: let's change it globally

packages/accordion/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "bl-accordion",
+  "name": "@blaze-elements/bl-accordion",
```
devopsmariocom: please update template as well

webpack.config.js
```diff
@@ -34,21 +34,21 @@ module.exports = ( env ) => {
         'test-helpers': './test-helpers.ts'
       },
       (env.element ? {
-        'main': `./packages/${ env.element }/index.ts`,
+        'index': `./packages/${ env.element }/index.ts`,
+        'index-with-dependencies': `./packages/${ env.element }/index.with.dependencies.ts`,
```
devopsmariocom: we should build this only for production

webpack.config.js
```diff
@@ -34,21 +34,21 @@ module.exports = ( env ) => {
...
         'main.demo': `./packages/${ env.element }/index.demo.tsx`,
-        'vendors': './vendors.ts',
         'polyfills': './polyfills.ts',
         'styles': './styles.ts'
       } : {
-        'main': './packages/index.ts',
+        'index': './packages/index.ts',
+        'index-with-dependencies': './packages/index.with.dependencies.ts',
```
devopsmariocom: same as 👆

devopsmariocom (changes requested):

packages/accordion/package.json
hopkins-tk: Updated...

package.json
```diff
@@ -21,7 +21,7 @@
     "test:watch:firefox": "concurrently \"npm run build:test:watch\" \"WCT_INDEX=./index.watch.test.html toolbelt/test.sh -p -l firefox\"",
     "build": "rimraf dist && npm run packagelist && webpack --env.prod",
     "build:test": "rimraf dist && npm run packagelist && webpack --env.test",
-    "build:test:watch": "npm run packagelist && webpack-dev-server --env.test",
+    "build:test:watch": "npm run packagelist && webpack-dev-server --env.test --host 0.0.0.0",
```
devopsmariocom: please just add this to webpack.config.js and remove all `--host 0.0.0.0`

```
    devServer: {
      host: '0.0.0.0'
    }
```

devopsmariocom (changes requested):

webpack.config.js
```diff
@@ -33,22 +33,26 @@ module.exports = ( env ) => {
         ),
         'test-helpers': './test-helpers.ts'
       },
-      (env.element ? {
-        'main': `./packages/${ env.element }/index.ts`,
+      (env.element ? ifProd({
```
devopsmariocom: I really don't want to be picky but we should create some helper function for this because it's really messy. If you want we can pair to come with some viable solution.

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "blaze-elements",
+  "name": "@blaze-elements/blaze-elements",
```
devopsmariocom: just one small thin can we keep main package unscoped? I don't think it's necessary.

webpack.config.js
```diff
@@ -20,35 +20,56 @@ const { getIfUtils, removeEmpty } = require( 'webpack-config-utils' );
...
+      'main.demo': _packagePrefix() + '/index.demo.ts',
+      'polyfills': './polyfills.ts',
+      'styles': './styles.ts'
+    };
+  }
+
+  function _entryPoint() {
+    if ( ifTest() ) {
+
+      return _testModules();
```
devopsmariocom: sorry I didn't realize that earlier but we should move returns from  _devModules, _prodModules and _testModules directly here so it's more straightforward.

Thanks

devopsmariocom (changes requested):

package.json
hopkins-tk: Is it really wise to have main package un-scoped & all sub-packages scoped?
- It does not make a lot of sense...

devopsmariocom (changes requested):
sorry just realised that this PR will destroy our showcase site. Let's discuss how we can easily enable showcase site within this PR without introducing big changes

webpack.config.js
```diff
@@ -145,34 +156,44 @@ module.exports = ( env ) => {
...
 
     ]),
     performance: {
       hints: ifTest() ? false : 'warning'
     }
   }
+
+
+  function ifProdOrSite( callback ) {
+    return ifProd( callback, ifSite( callback ) );
```
devopsmariocom: wow looks like you have tried to open gate to callback hell dimension 😄 

mr-clippy

webpack.config.js
```diff
@@ -145,34 +156,44 @@ module.exports = ( env ) => {
       }),
```
devopsmariocom, line 134: please add ifProdOrSite here as well but without true false hell and add no callback support to the helper

devopsmariocom (changes requested):

devopsmariocom (approved):

webpack.config.js
```diff
@@ -138,41 +149,63 @@ module.exports = ( env ) => {
...
       }))
 
     ]),
     performance: {
       hints: ifTest() ? false : 'warning'
     }
   }
+
+
+  function ifProdOrSite( callback = true ) {
```
Hotell: why is argument with name `callback` boolean?

webpack.config.js
```diff
@@ -57,6 +65,9 @@ module.exports = ( env ) => {
       extensions: [ '.js', '.ts', '.tsx' ]
     },
 
+    devServer: {
```
Hotell: why is this needed? localhost is used by default

webpack.config.js
```diff
@@ -138,41 +149,63 @@ module.exports = ( env ) => {
...
-          excludeChunks: [ 'main' ], // Exclude 'main' as it is included in 'main.demo'
+          excludeChunks: [ 'index', 'index-with-dependencies' ], // Exclude 'index' & 'index-with-dependencies' as it is included in 'main.demo'
           inject: 'head',
-          chunksSortMode: buildChunksSort([ 'polyfills', 'vendors', 'styles', 'main', 'main.demo', 'test-helpers', 'test' ])
+          chunksSortMode: buildChunksSort([ 'polyfills', 'styles', 'index', 'index-with-dependencies', 'main.demo', 'test-helpers', 'test' ])
       }))
 
     ]),
     performance: {
       hints: ifTest() ? false : 'warning'
```
Hotell: you didn't write this, but it's really necessary only for production build. We don't need to get bugged all the time in both console and browser devtools that unminified files are exceeding some file size. pls change to
`hints: ifProd() && 'warning'`

toolbelt/templates/package.json
```diff
@@ -1,10 +1,10 @@
 {
-  "name": "bl-generic",
+  "name": "@blaze-elements/generic",
```
Hotell: `generic` is not very "generic" IMHO. Let's use a placeholder style `@blaze-elements/<component-name>`

packages/typography/package.json
```diff
@@ -1,10 +1,10 @@
 {
-  "name": "bl-typography",
+  "name": "@blaze-elements/typography",
   "version": "1.2.0",
   "main": "dist/main.js",
   "license": "MIT",
   "scripts": {
-    "start": "../../node_modules/.bin/webpack-dev-server --config ../../webpack.config.js --env.dev --hot --host 0.0.0.0 --env.element=typography",
+    "start": "../../node_modules/.bin/webpack-dev-server --config ../../webpack.config.js --env.dev --hot --env.element=typography",
```
Hotell: this is a nit pick but better stay consistent.

Let's use webpack `env` flags together, then `wp-devServer` flags.

```diff
-"start": "../../node_modules/.bin/webpack-dev-server --config ../../webpack.config.js --env.dev --hot --env.element=typography"
+"start": "../../node_modules/.bin/webpack-dev-server --config ../../webpack.config.js --env.dev --env.element=typography --hot"

```

styles.ts
```diff
@@ -1,2 +1,3 @@
-import 'style-loader!css-loader!sass-loader!blaze/scss/blaze.scss';
+import 'style-loader!css-loader!sass-loader!blaze/scss/generics.global.scss';
```
Hotell: Why is this done this way? 
import are supported directly within `scss`

```scss
// style.scss
@import "~blaze/scss/generics.global.scss";
@import "~blaze/scss/components.typography.scss";
```

```ts
// style.ts
import './styles.scss';
```

webpack.config.js
```diff
@@ -18,37 +18,45 @@ const { getIfUtils, removeEmpty } = require( 'webpack-config-utils' );
 
 
 module.exports = ( env ) => {
-  const { ifProd, ifNotProd, ifTest, ifNotTest } = getIfUtils( env );
+  const { ifProd, ifNotProd, ifTest, ifDev, ifSite } = getIfUtils( env, ['prod', 'test', 'dev', 'site'] );
+
+  function _packagePrefix() {
```
Hotell: there is no reason this to be a function.

rather use const with proper path resolve and more descriptive name

```diff
-function _packagePrefix() {
-  return './packages' + ( env.element ? `/${env.element}` : '' );
-}
+const packagePath = resolve('./packages', env.element || '' );
```

webpack.config.js
```diff
@@ -18,37 +18,45 @@ const { getIfUtils, removeEmpty } = require( 'webpack-config-utils' );
...
+    return './packages' + ( env.element ? `/${env.element}` : '' );
+  }
+
+  function _entryPoint() {
```
Hotell: again nit pick.

name functions by what they do.

This one returns entry point, so `function getEntryPointConfig() {` is definitely more appropriate.

Points to do:
- properly resolve paths via `resolve` instead of string concat
- you dont need to use strings for object keys

Points to consider:
- also we could make it pure and extract it from direct config

```js
function getEntryPointConfig(basePath, {isTest,isProd}={}) {
   if( isTest ){
     return {
       test: resolve(basePath, '/index.test.ts'),
       'test-helpers': './test-helpers.ts',
     };
   }
   if( isProd ){
      return /*...*/
   }
   return /*...*/

   
}
```

then:

```js
module.exports = ( env ) => {
  const { ifProd, ifNotProd, ifTest, ifDev, ifSite } = getIfUtils( env, ['prod', 'test', 'dev', 'site'] );
  const packagePath = resolve('./packages', env.element || '' );
  const entry = getEntryPointConfig(packagePath,{isTest:isTest(),isProd:isProd()});
  // ...
}
```

webpack.config.js
```diff
@@ -18,37 +18,45 @@ const { getIfUtils, removeEmpty } = require( 'webpack-config-utils' );
...
+        'test': _packagePrefix() + '/index.test.ts',
+        'test-helpers' : './test-helpers.ts'
+      };
+
+    }
+
+    if ( ifProd() ) {
+
+      return {
+        'index': _packagePrefix() + '/index.ts',
```
Hotell: **NOTE** 🚨

this build entry points are not doing what you think they are doing.

`index.ts` imports `define` from skate, so as a end result. skate will be always bundled. This is not what we wanted.

What you need to do:

- for prod build you have to provide `externals` for main index.ts packages 
```
externals: {
  skate: 'skatejs'
}
```
- for prod build of `index-with-deps.ts` that vendor import is completely unnecessary and may increase bundle size. We are using three shaking, so only used functions/imports are bundled. only thing that you need to import there are `polyfills`

Hotell (changes requested):

packages/tree/package.json
```diff
@@ -1,10 +1,10 @@
 {
-  "name": "bl-tree",
+  "name": "@blaze-elements/tree",
   "version": "1.2.0",
   "main": "dist/main.js",
```
Hotell: there is no `main.js` generated during prod build, pls change in all packages to `"main": "dist/index.js"`

Hotell (changes requested):

webpack.config.js
hopkins-tk: Because **localhost** is **127.0.0.1** and it is useless when you need to test from e.g. VM or when you are running it in VM and want to test from outside.

Typical use case is testing in IEs.

styles.css
```diff
@@ -1,3 +1,6 @@
+@import "~blaze/scss/generics.global.scss";
```
Hotell: does this work? it should be `styles.scss`

webpack.config.js
```diff
@@ -181,33 +156,59 @@ module.exports = ( env ) => {
...
     ]),
     performance: {
-      hints: ifTest() ? false : 'warning'
+      hints: ifProd() && 'warning'
     }
   }
 
 
-  function ifProdOrSite( callback = true ) {
+  function ifProdOrSite( returnValue = true ) {
```
Hotell: ok now I see what you try to do in here - mimic `getIfUtils`.
So we should adhere to their implementation and make things easier. 

```js
function getCustomIfUtils({ifDev, ifProd, ifSite}={}){
   return {
     ifProdOrSite(value, alternate){  return getByEnvValue(ifProd() || isSite(), value, alternate )  },
     ifDevOrSite(value, alternate){ return getByEnvValue(ifDev() || isSite(), value, alternate ) }
   }

   function getByEnvValue(envValue, value, alternate ){
     return isUndefined(value) ? envValue : propIf(envValue, value, alternate )
   }

   function isUndefined(val){ return typeof val === "undefined" }
   function getValue(val) {
    return JSON.parse(val)
   }
   function propIf(add, value, alternate) {
     return getValue(add) ? value : alternate
   }
}
```

and then on the top

```js
module.exports = ( env ) => {
  const { ifProd, ifNotProd, ifTest, ifDev, ifSite } = getIfUtils( env, ['prod', 'test', 'dev', 'site'] );
  const { ifProdOrSite, ifDevOrSite } = getCustomIfUtils({ifDev,ifProd,ifSite});
  // ...
}
```

packages/tree/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@blaze-elements/tree",
   "version": "1.2.0",
-  "main": "dist/main.js",
+  "main": "dist/index.js",
```
Hotell: not sure, if it doesn't have to be `dist/index.min.js` ?

Hotell (changes requested):

webpack.config.js
Hotell: can you add comment there, because it's not obvious why is it there. thanks

Hotell:
### Things to do 

> ( Im putting it here because stupid github has hidden those comments )

this build entry points are not doing what you think they are doing.

`index.ts` imports `define` from skate, so as a end result. skate will be always bundled. This is not what we wanted.

What you need to do:

- [ ] for prod build you have to provide `externals` for main index.ts packages 
```
externals: {
  skatejs: 'skatejs'
}
```
- [ ] for prod build of `index-with-deps.ts` that vendor import is completely unnecessary and may increase bundle size. We are using three shaking, so only used functions/imports are bundled. only thing that you need to import there are `polyfills`

this will be addressed in another PR 
https://github.com/wc-catalogue/blaze-elements/issues/177

webpack.config.js
```diff
@@ -18,37 +18,20 @@ const { getIfUtils, removeEmpty } = require( 'webpack-config-utils' );
 
 
 module.exports = ( env ) => {
-  const { ifProd, ifNotProd, ifTest, ifNotTest } = getIfUtils( env );
+  const { ifProd, ifNotProd, ifTest, ifDev, ifSite } = getIfUtils( env, ['prod', 'test', 'dev', 'site'] );
+
+  const packagePath = resolve( './packages', env.element || '' );
```
Hotell: this doesn't work. Pls don't always copy paste what I write in review :D Im not unit tested because I'm human :D 

there is context needed, again we can make things DRY

```js
const context = resolve(__dirname);
const packagePath = resolve( context, './packages', env.element || '' );

return {
   context,
   entry: getEntryPointConfig( packagePath, { ... } ),
  // ...
   
```

packages/typography/package.json
hopkins-tk: I am not sure what you meant by that.

Hotell: isn't the diff obvious? 

`--hot` is argument for webpack-dev-server
`--env.<foo>` is for webpack

webpack.config.js
hopkins-tk: If we add **externals**, skate will not make it to any of the bundles, as externals are global, or at least I have not found anything else in Webpack "documentation".

As far as polyfills goes - I think we did not want to ship them at all.

Hotell: well, because the config is wrong written. Next time don't hesitate to discuss things with me. @elmariofredo is not the only devstack "guru" in the room ;) :P Anyway last 2 devstacks that are open source were build primarily by me :P just saying ☮

here is what we need to do:
https://webpack.js.org/concepts/configuration/#multiple-targets

webpack.config.js
```diff
@@ -160,22 +163,45 @@ module.exports = ( env ) => {
     }
   }
 
+  function getCustomIfUtils( { ifDev, ifProd, ifSite } = {} ) {
 
-  function ifProdOrSite( returnValue = true ) {
+    return {
+
```
Hotell: for the next time, we don't need to produce 1000 lines of blank lines, mario 200px font era is long gone. 

Also it's a smell if you need lot of blank lines in the function to have separated content, it probably does too much and should be chunked to separate small footprint functions

hopkins-tk: Not the right place to discuss it, but we should probably discuss how we want the code to be formatted to have it unified...

I was not terribly happy with new lines, but got used to put them everywhere and not aware that we have change it - maybe I did not pay attention :)

So lets just make some rules what we deffinately want and what we do not care about, because e.g. the formatter seams to be switching it one way and than the other constantly in certain files :)

Hotell: Yup yup. That's why I wrote "next time". Formatter had a bug with ending spaces in curly braces which I would like to also removr, the code looks wierd and long with all those spaces. We will discuss is in person for sure. Thanks!

Hotell:
Great job @hopkins-tk will take a look tomorrow. If everything will be OK, I'll merge it. Thx!

Hotell (approved):

PR #154, opened 2017-02-03
feat(app-layout): add

devopsmariocom:


packages/_helpers/match-media.ts
```diff
@@ -0,0 +1,23 @@
+
+/**
+ * On media change call callback with matches argument
+ */
+export function matchMedia( mediaQuery: string, callback: ( matches: boolean ) => void ): () => void {
+
+  function handleOrientationChange( mql: MediaQueryList ) {
```
hopkins-tk: Is it really `handleOrientationChange`?

Seems more like something that will be invoked when a media query changes.

packages/_helpers/match-media.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  }
+
+  const mql = window.matchMedia( mediaQuery );
+  mql.addListener( handleOrientationChange );
+
+  // Call manually to get initial state
+  handleOrientationChange( mql );
+
+  return () => {
```
hopkins-tk: Just something to think about...

Are those curly brackets really necessary?

Wouldn't following be better as it is oneliner?
```
return () => mql.removeListener( handleOrientationChange );
```

packages/app-layout/AppLayout.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+import { HamburgerButton } from './components/HamburgerButton';
+import { Drawer } from './components/Drawer';
+import { Overlay } from './components/Overlay';
+import { Nav } from './components/Nav';
+import { NavContent } from './components/NavContent';
+
+import styles from './AppLayout.scss';
+
+const DEFAULT_RESPONSIBLE_WIDTH = '640px';
```
hopkins-tk, line 15: Just to think about:
- is it really good to have user (when specifying `responsiveWidth`) use string with some units?
- wouldn't it be better to have the width as a number and just always use **px**?

packages/app-layout/AppLayout.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+    this.setupQueryMatchListener();
+
+  }
+
+  disconnectedCallback() {
+
+    super.disconnectedCallback();
+
+    this.matchMediaUnsubscribe();
```
hopkins-tk, line 101: I would probably swap those two lines just to be safe:

```
- connect PARENT
-- connect SELF
-- ...
-- disconnect SELF
- disconnect PARENT
```

packages/app-layout/AppLayout.tsx
```diff
@@ -0,0 +1,161 @@
...
+      } ),
+      responsiveWidth: prop.string( {
+        initial: DEFAULT_RESPONSIBLE_WIDTH,
+        attribute: {
+          source: true
+        }
+      } ),
+      narrow: prop.boolean( {
+        attribute: {
+          source: true,
```
hopkins-tk: Is it something we want user to be configuring?

Isn't it just local property, which is being set in `matchMedia(...)`

packages/app-layout/AppLayout.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+export class AppLayout extends Component<AppLayoutProps> {
+
+  static get is() { return 'bl-app-layout'; }
+
+  static get props() {
+    return {
+      drawerVisible: prop.boolean( {
+        attribute: {
+          source: true
```
hopkins-tk: Is it something we want user to be configuring?

Isn't it just local property, which is being set in `openDrawer()` / `closeDrawer(...)`

packages/app-layout/AppLayout.tsx
```diff
@@ -0,0 +1,161 @@
...
+    this.matchMediaUnsubscribe();
+
+    this.matchMediaUnsubscribe = matchMedia(
+      `(min-width: ${this.responsiveWidth})`,
+      ( matches ) => {
+
+        if ( matches ) {
+
+          this.openDrawer();
+          this.narrow = false;
```
hopkins-tk: This condition could be probably slightly simplified as following:

1) moving the setting of `narrow` up and make it one liner:
```
this.narrow = !matches;
```

2) even the rest could be than squashed to a single line:
```
matches ? this.openDrawer() : this.closeDrawer();
```

packages/app-layout/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "bl-app-layout",
+  "version": "1.0.0",
+  "main": "dist/main.js",
+  "license": "MIT",
+  "scripts": {
+    "start": "../../node_modules/.bin/webpack-dev-server --config ../../webpack.config.js --env.dev --hot --host 0.0.0.0 --env.element=app-layout",
+    "test": "npm run build:test && ../../node_modules/.bin/wct -l firefox",
```
hopkins-tk: `firefox` :)

I would probably regenerated it after rebase completely as I think we have been adding some things to the template...
There is also awaiting https://github.com/wc-catalogue/blaze-elements/pull/202

packages/drawer/Drawer.tsx
```diff
@@ -22,28 +24,41 @@ export class Drawer extends Component<DrawerProps> {
...
+      floating: prop.boolean( {
+        attribute: {
+          source: true
+        },
+        initial: true
       } )
     };
   }
 
   position = 'top';
```
hopkins-tk, line 44: I know, this is unrelated, but could you please move the initial value to `prop` definition as you have it for `floating`.

Thx ;-)

packages/index.demo.tsx
```diff
@@ -0,0 +1,22 @@
+import './accordion/index.demo';
```
hopkins-tk: You might not have rebased - this has been moved to **packages/blaze-elements**.

packages/index.test.ts
```diff
@@ -0,0 +1,22 @@
+import './accordion/index.test';
```
hopkins-tk: You might not have rebased - this has been moved to **packages/blaze-elements**.

packages/index.ts
```diff
@@ -0,0 +1,22 @@
+import './accordion/index';
```
hopkins-tk: You might not have rebased - this has been moved to **packages/blaze-elements**.

packages/nav/index.demo.tsx
```diff
@@ -1,7 +1,7 @@
 import { h, Component } from 'skatejs';
 import { Nav } from './Nav';
-import { NavItem } from './Nav-item';
-import { NavContent } from './Nav-content';
+import { NavItem } from './NavItem';
+import { NavContent } from './NavContent';
```
hopkins-tk: All the 3 imports should be from **index** now, as you have added it there ;-)

So:
```
import { Nav } from './index';
import { NavItem } from './index';
import { NavContent } from './index';
```

hopkins-tk (changes requested):

hopkins-tk:
I know we have agreed on the slotted version, but as it can be seen now it requires following components to be bundled with app layout component:
- Drawer
- Nav
- NavContent
- Overlay

So it might be something we should reconsider in the next iteration...

global.d.ts
```diff
@@ -129,3 +130,16 @@ interface Document {
 // we are not using Chai anymore. We use Expect which has Jest style assertions
 // @TODO remove this once all test are refactored to use `epxect` instead of chai
 declare var expect: Chai.ExpectStatic;
+
+// TODO: Use proper types
+declare namespace JSX {
```
hopkins-tk: Please try to discuss with @Hotell how to tackle this as I recall some issues he was mentioning...

packages/drawer/Drawer.tsx
Hotell: Also pls don't use `initial`, this is deprecated and will be removed in next version of skate. If you have any questions about skate API don't hesitate to ask me ( as I am core contributor and know internals and what not )

packages/app-layout/AppLayout.tsx
devopsmariocom: I'm following polymer [way](https://github.com/PolymerElements/app-layout/blob/master/app-drawer-layout/app-drawer-layout.html#L172) and yes you don't always want to use px let's let developer to choose.

packages/app-layout/AppLayout.tsx
devopsmariocom: I'm following [documentation](https://github.com/skatejs/skatejs/blob/e2eaa330a34ce790bd515fa6c0d02bd7c9e23e62/docs/examples/README.md#L31) and according to [code](https://github.com/skatejs/skatejs/blob/f5ab7073f91e89003b46303e50e2f9ce3c3b8163/src/api/component.js#L221) I don't see any harm in that. Please explain and let's submit PR to skate doc.

packages/app-layout/AppLayout.tsx
devopsmariocom: it depend on style and preference for me line `this.narrow = !matches;` doesn't have much sense, I feel that my if statement is more explicit. In general negating value is always harder to grasp for our simple minds 😉 .

packages/drawer/Drawer.tsx
devopsmariocom: @Hotell please be so kind and share link to deprecation notice in skate doc/code, all I see is this https://skatejs.gitbooks.io/skatejs/content/docs/api/Component.html#initial

Thanks

Hotell: I really don't have time to reference everything to everyone, you can follow my advice as person who knows the internals of the library or feel free to google it 4 yourself ( good start is  project github issues and PR's ). Once again It is going away in next iteration of skate + it is bad design pattern to use ( when setting initial prop values we use `default`, we already talked about that within team as you may remember )

packages/app-layout/AppLayout.tsx
devopsmariocom: developer might want to open drawer programatically

packages/app-layout/AppLayout.tsx
devopsmariocom: you might want to at least let developer to read narrow property value, then I was following future default attribute value {source: true}. I was trying to set source to false but got error `Uncaught Error: attribute 'source' or 'target' is missing.`.

devopsmariocom:
@hopkins-tk please can you expand your thoughts on bundling other elements? I thought that we did discuss this issue and agreed that this won't be problem when developer will use `blaze-elements` bundle where it should get deduplicated.

packages/app-layout/AppLayout.tsx
Hotell: Hint:
Setting source: false doesn't make any sense. If you don't wanna use attribute reflection don't use attribute at all.

devopsmariocom: ah you are right I kind of thought that attribute: { source: true } will be default for all pops, but it will be default only for attribute: true, ok will change it to not use attribute reflection at all

Hotell: I haven't read your code, but thing to consider ( I've already discussed it with Dave. )
Do we really need this to be declared via prop API ( it makes only sense if you wanna really trigger re render. ) If don't rather use it directly on instance and if you are mutating props that are defined via static props get API, always use props ( which is setState equivalent in React ).

Also when setting via props function updates are batched so they are predictable. Instead of you set directly to this, it is asynchronous operation. In skate we are going to strictly enforce using props in next release.  Don't hasitate to ask further. Cheers!

Hotell: Yup that's right

devopsmariocom: > I haven't read your code, but thing to consider ( I've already discussed it with Dave. ) Do we really need this to be declared via prop API ( it makes only sense if you wanna really ? trigger re render. ) If don't rather use it directly on instance and if you are mutating props that are defined via static props get API, always use props ( which is setState equivalent in React ).

yep that's the case, I need to rerender when this property change.

> Also when setting via props function updates are batched so they are predictable. Instead of you set directly to this, it is asynchronous operation. In skate we are going to strictly enforce using props in next release. Don't hasitate to ask further. Cheers!

ups did miss this pattern, will fix that. If I may I would suggest to rather change naming to either setProps or even setState as props feels too ambiguous.

Hotell: Well it's not, because props is a getter or setter ( depends on argument arity)

devopsmariocom: I like code to be more explicit(like separating to get/setProps) as arity is not that obvious from first time reader point of view also seems even tray use it with different naming https://github.com/skatejs/skatejs/blob/7db2988c4d6c20651e8dad2145912ee3bea659e4/src/api/component.js#L22 :). 

Anyway that's just nitpicking, I'm bit confused from asynchronicity of props. As you are very well aware of setAsyncState in react does this count for skate as well? What I have got from your note, it's async when done directly, but using props setter is sync. Does it mean that we can count on that and use this.someProp as getter (as we are suppose to use props setter every time) or we should rather use pops getter as well? Are there any cases where we should rather use async setter?

Hotell: Using props as getter doesn't make much sense. I don't care what naming is used internally in skate :) props is public API. Deal with it. We are going to make props behave like setState so it can accept asynchronous FN callback and make itself asynchronous  but via batches ( as react does ) .

So as a summary. Always set public API ( defined via public API - static get props / till we have decorators ready  ) via props FN. Internal private variables via direct this.foo=123

devopsmariocom: thanks for clarification didn't get props getter either

packages/app-layout/AppLayout.tsx
hopkins-tk: Ok :-)

packages/app-layout/AppLayout.tsx
```diff
@@ -0,0 +1,157 @@
...
+  private setupQueryMatchListener() {
+
+    // Unsubscribe previously subscribed listener
+    this.matchMediaUnsubscribe();
+
+    this.matchMediaUnsubscribe = matchMedia(
+      `(min-width: ${this.responsiveWidth})`,
+      ( matches ) => {
+
+        props( this, {
```
hopkins-tk: Nice ;-)

packages/app-layout/AppLayout.tsx
hopkins-tk: Ok in that case, aren't we immediately overriding it due to `setupQueryMatchListener` & `matchMedia`?

packages/app-layout/AppLayout.tsx
hopkins-tk: Ok, if it is recommended by skate...

I was going with the general feeling:
- let parent create/bootstrap what is needed, than enrich it
-- I might need intimated object
- when cleaning/destroying, clean what I did, than let parent clean his stuff
-- if parent's cleaning process would contain destroying the context somehow, to clean the mess, that might have been created
-- in other words, I might still need fully initiated object, when I am doing my cleanup...

packages/drawer/Drawer.tsx
hopkins-tk: @Hottel r eminded me off the `default` and we are already using it, so it would be better to use it to be consistent.

I missed that at first - sorry...

hopkins-tk:
@elmariofredo you are right - that's why I just put it so that we can discuss it, but leave it now as is...

I am not worried about size of the produced bundle, but the number of necessary classes/files in the source code...

But again for the fist iteration it is fine.

package.json
```diff
@@ -9,7 +9,7 @@
     "ts:style": "npm run ts:lint:fix && npm run ts:format",
     "start": "webpack-dev-server --env.dev --env.element=blaze-elements --hot",
     "start:prod": "pushstate-server dist/",
-    "test": "npm run build:test && toolbelt/test.sh",
+    "test": "npm run build:test && toolbelt/test.sh -l firefox",
```
hopkins-tk: "firefox"? :)

hopkins-tk (approved):

PR #156, opened 2017-02-03
refactor(calendar): init component logic & repair range fn & tests

davidrus:
affects: bl-calendar

Some notes:
- month and year must be defined in `prop()` because we need to rerender after any change
  - I tried to use `props(...)`, but I don't wanna have year and month in type `Props` because I can't have public interface for setting this 
- I replaced setter of `selectedDate` to use `props(...)`
- new tests including test of i18n

### DIFF
--- a/packages/calendar/Calendar.test.tsx
+++ b/packages/calendar/Calendar.test.tsx
@@ -1,11 +1,12 @@
 import * as expect from 'expect';
 import * as isSameDay from 'date-fns/is_same_day';
 import * as parse from 'date-fns/parse';
-import { Calendar } from './';
+import { Calendar } from './index';
 import { CalendarButton } from './components/Button';
 
 import { h, mount, WrappedNode } from 'bore';
 import { emit } from 'skatejs';
+import { GenericEvents } from '../_helpers';
 
 describe( Calendar.is, () => {
 
@@ -194,7 +195,7 @@ describe( Calendar.is, () => {
 
         let changeTriggered = false;
         let selectedDate: Date;
-        const handleChange = ( e: CustomEvent ) => {
+        const handleChange = ( e: GenericEvents.CustomChangeEvent<Date> ) => {
           changeTriggered = true;
           selectedDate = e.detail.value;
         };
--- a/packages/calendar/Calendar.test.tsx
+++ b/packages/calendar/Calendar.test.tsx
@@ -1,5 +1,6 @@
 import * as expect from 'expect';
 import * as isSameDay from 'date-fns/is_same_day';
+import * as parse from 'date-fns/parse';
 import { Calendar } from './';
 import { CalendarButton } from './components/Button';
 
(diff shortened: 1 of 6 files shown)


davidrus:
Rebased

davidrus:
Ready to review @Hotell

packages/calendar/Calendar.test.tsx
```diff
@@ -1,17 +1,223 @@
-import { Calendar } from './Calendar';
+import * as expect from 'expect';
+import * as isSameDay from 'date-fns/is_same_day';
+import * as parse from 'date-fns/parse';
+import { Calendar } from './';
```
Hotell: we have to import explicitly from './index' , because some issues with build....

packages/calendar/Calendar.test.tsx
```diff
@@ -1,17 +1,223 @@
...
+
+    describe( `(dateChange)`, () => {
+
+      it( `should emit onChange event with selected day value`, () => {
+
+        const expectedDate = parse( '1987-12-09' );
+
+        let changeTriggered = false;
+        let selectedDate: Date;
+        const handleChange = ( e: CustomEvent ) => {
```
Hotell: nit pick. use our customChangeEvent type

packages/calendar/Calendar.tsx
```diff
@@ -61,14 +86,14 @@ export class Calendar extends Component<CalendarProps> {
   }
 
   static range( n: Number ) {
```
Hotell: we should use `number` === primitives `Number` is object

packages/calendar/index.demo.tsx
```diff
@@ -1,7 +1,6 @@
 import { h, Component, prop } from 'skatejs';
-import { Calendar, CalendarChangeEvent } from './Calendar';
-
-
+import { Calendar } from './index';
+import { CustomChangeEvent } from '../_helpers/definitions/events';
```
Hotell: import only from "barel", always double check what Webstorm did to you! 🗡 :D

packages/calendar/Calendar.tsx
```diff
@@ -193,15 +227,19 @@ export class Calendar extends Component<CalendarProps> {
...
-    this.selectedDate = parse( this.selectedDate );
     this.year = getYear( this.selectedDate );
     this.month = getMonth( this.selectedDate );
   }
 
   private initDays() {
     const date = new Date( this.year, this.month );
     const days: Date[] = [];
-    let currentDate = startOfWeek( date, { weekStartsOn: this.weekStartsOn === WEEK_STARTS_ON ? 0 : 1 });
+    let currentDate = startOfWeek( date,
```
Hotell: let's talk about this and the following double forEach

packages/calendar/Calendar.tsx
```diff
@@ -61,14 +86,14 @@ export class Calendar extends Component<CalendarProps> {
   }
 
   static range( n: Number ) {
-    return Array.from( Array( n ).keys() );
+    return Array.from( { length: n } as any, ( value: number, key: number ) => key );
```
Hotell: hmm what is this doing?

Hotell (changes requested):
Good job overall ! just few tweaks

Hotell (approved):

PR #157, opened 2017-02-03
refactor: add latest bore and allow ts declaration generation

Hotell:


packages/button/Button.tsx
```diff
@@ -1,33 +1,44 @@
-import { h, Component, prop } from 'skatejs';
+import { h, Component, prop, ComponentProps } from 'skatejs';
 
-import { ColorType, cssClassForColorType, css } from '../_helpers';
+import {
+  ColorType,
```
davidrus, line 4: Just thinking - what about change order of imports to first helpers, then types? like:
`import {
 cssClassForColorType, 
 css, 
 ColorType, 
 IntrinsicCustomElement, 
 IntrinsicBoreElement, 
 ClickEvent 
} from '../_helpers'`

packages/button/Button.test.tsx
```diff
@@ -1,45 +1,64 @@
...
-      expect( element.has( '.c-button' ) ).toBe( true );
+
+    it( `should render via JSX IntrinsicElement`, () => {
+      return mount( <bl-button>Hello</bl-button> ).wait(( element ) => {
+        expect( element.node.localName ).toBe( Button.is );
+      });
+    });
+
+    it( `should render`, () => {
+      return mount( <Button /> ).wait(( element ) => {
```
devopsmariocom, line 22: shouldn't we use `<bl-button>` here? I know that it's stated here https://github.com/wc-catalogue/blaze-elements/blob/master/docs/styleguides/TESTING.md#example--button- but I did forgot why.

packages/button/Button.tsx
Hotell: hmm thats too much nit picking IMHO. But I see your point. We are importing a lot of stuff from `helpers` but that's what is it for.

Important is to always order `imports` in order 

- 3rd party
- local ( non package related ) - this should go away when we have proper build 
- package internal relative imports


Maybe we can extract this definitions to root `/packages/definitions.ts`

Also that `_helpers` folder is wrong, it should be separate package named `core` or `base`

davidrus: Maybe next time 👍

davidrus (approved):

packages/button/Button.tsx
```diff
@@ -1,33 +1,44 @@
...
+  ClickEvent
+} from '../_helpers';
 
 import styles from './Button.scss';
 
-// public
-type ButtonProps = Props & EventProps;
-type Props = {
+export type ButtonProps = Props & Attrs & EventHandlers;
+export type Attrs = {
```
devopsmariocom: we could probably have everything in `Props` as all props will have `attribute:{source: true}` as default right? Or is there any reason to explicitly state it.

packages/button/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "1.2.0",
   "main": "dist/main.js",
   "license": "MIT",
+  "typings": "definitions/index.d.ts",
```
devopsmariocom, line 6: please add it also to template https://github.com/wc-catalogue/blaze-elements/blob/master/toolbelt/templates/package.json

packages/button/tsconfig.json
```diff
@@ -1,5 +1,9 @@
 {
   "extends": "../../tsconfig.json",
+  "compilerOptions": {
```
devopsmariocom, line 3: same as 👆

packages/button/Button.tsx
Hotell: cheers mate 🍺

packages/button/package.json
Hotell: this is up to discussion, because current build is non existent till https://github.com/wc-catalogue/blaze-elements/pull/153

packages/button/Button.tsx
Hotell: it depends on Component ofc.
 Button "primary" api is via attributes, that is also what is tested in test. Also this "clearly" separates type definitions to common pattern.
`Props` - primary API, should be preferred over attributes whenever possible
`Attrs` - only if primary API is via attributes,
`Events` - event aliases which we emit, specifically for bore `h`
`EventHandlers` - event aliases which we emit, specifically for IntrinsicElement consumption within other Virtual Dom view renderers

Attrs and Events are specifically for bores `h` while intersection of all 3 is our intrinsic element API.

We can discuss it further ofc, this is just what I ended up with, after evaluation of all possible scenarios

packages/button/Button.test.tsx
Hotell: > shouldn't we use here? 

use what?

packages/button/package.json
devopsmariocom: well template should reflect desired state for each package, also it doesn't have to do anything with our current build if definitions are not working it shouldn't be here as well.

Simply I want to have package.json to be auto generated to some degree so we can easily change tooling for all packages.

packages/button/Button.test.tsx
devopsmariocom: ups sorry forgot back ticks around the tag

packages/button/Button.tsx
devopsmariocom: yes I got the idea but I'm not sure what does it mean

> only if primary API is via attributes

in reality. How you decide that that particular property is primarily attribute? Also I though that we won't test attributes as that's functionality which is tested in skatejs and in future it will be by default with `attribute:{source: true}`.

Let's discuss this with rest directly wdyt @hopkins-tk @davidrus?

packages/button/package.json
Hotell: My comment was about the path of typings within package.json

devopsmariocom: as you know there is build process from very beggining but its building test files instead of production bundles to dist folder, not sure how this is related to #153 if it is related somehow  then you should wait before #153 is closed and even then add it to the template as well

hopkins-tk (approved):

packages/button/Button.test.tsx
devopsmariocom: 👋 still waiting for an answear

devopsmariocom (changes requested):

packages/button/Button.test.tsx
Hotell: don't remember either ¯\_(ツ)_/¯ 

maybe few suggestions about your review attitude:
- if you intend to stop merging things because some unrelated unanswered questions, that's very unfortunate for velocity and for everyone that's participating on this project
- we can change it ofc in another PR which should refactor other tests that have the same pattern and update the docs accordingly

devopsmariocom: then please fix it to `<bl-button>`

thanks

p.s. this tone of suggestion is really unfortunate, point of PR is answering questions. PR comments are not the right place for discussing attitudes and any PR unrelated topic. Please contact me directly if you want to discuss this further. Thanks

Hotell: Can you pls unblock merging of this PR. You are blocking rest of the team. thx

Hotell:
Closing in favor of #168 

https://github.com/wc-catalogue/blaze-elements/issues/167 should be addressed in separate PR

Issue #180, opened 2017-02-11
build: toolbelt templates


Hotell:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => You can submit support request here, but we preffer to chat about it realtime at our [ngParty slack](https://ngparty.herokuapp.com/)
```

**Current behavior**

Everytime we run toolbelt for generating templates it will override existing one, which will remove all dependencies that current package had...

**Expected behavior**

It should take into account every package.json dependencies and leave them as they are ( it should change only common config )

hopkins-tk:
Another related question - maybe more generic.

What are all the things that we want to override and what not?

PR #187, opened 2017-02-12
chore(generate-configs): prevent dependencies override

devopsmariocom:


devopsmariocom:
Closes #180

toolbelt/generate-configs.js
```diff
@@ -3,29 +3,43 @@
...
+
 for (let packageName of packageList) {
 
   console.log(packageName);
 
-  fs.createReadStream('./toolbelt/templates/tsconfig.json').pipe(fs.createWriteStream(`./packages/${packageName}/tsconfig.json`));
+  const packageRootDir = path.resolve(packagesRootDir, packageName);
+
+  // Copy tsconfig.json template to each package folder
+  fs.writeFileSync(path.resolve(packageRootDir, 'tsconfig.json'), JSON.stringify(tsconfigJsonTemplate, null, 2) + "\n")
```
Hotell, line 25: instead of using comments, a function with appropriate name could have been superior ;)

Hotell: everytime you concat strings with `+` a kitten dies :D

toolbelt/generate-configs.js
```diff
@@ -3,29 +3,43 @@
...
 const packageList = require('./packages-list.js').packageList();
 
 console.log('Generating package.json and tsconfig.json for:');
 
+const packagesRootDir = path.resolve(__dirname, '..', 'packages');
+const templatesRootDir = path.resolve(__dirname, 'templates');
+
+const packageJsonTemplate = require( path.resolve(templatesRootDir, 'package.json') );
+const tsconfigJsonTemplate = require( path.resolve(templatesRootDir, 'tsconfig.json') );
+
```
Hotell, line 17: this `for loop` feels like it's 1996 again :P

toolbelt/generate-configs.js
```diff
@@ -3,29 +3,43 @@
...
   console.log(packageName);
 
-  fs.createReadStream('./toolbelt/templates/tsconfig.json').pipe(fs.createWriteStream(`./packages/${packageName}/tsconfig.json`));
+  const packageRootDir = path.resolve(packagesRootDir, packageName);
+
+  // Copy tsconfig.json template to each package folder
+  fs.writeFileSync(path.resolve(packageRootDir, 'tsconfig.json'), JSON.stringify(tsconfigJsonTemplate, null, 2) + "\n")
 
-  const originalPackageJson = require(`../packages/${packageName}/package.json`);
+  // Generate package.json file base on package package.json file and template
```
Hotell, line 27: package package packaaaage, omelet du fromaaaaz :D sorry, can we rename it somehow ? thx

toolbelt/generate-configs.js
```diff
@@ -3,29 +3,43 @@
...
+  const packagePackageJsonPath = path.resolve(packageRootDir, 'package.json');
+
+  const originalPackageJson = (
```
Hotell, line 30: there is no need for parenthesis. We write ternary on new lines
```
const originalPackageJson = fs.existsSync( packagePackageJsonPath ) 
  ? require( packagePackageJsonPath ) 
  : {}
```

toolbelt/generate-configs.js
```diff
@@ -3,29 +3,43 @@
...
+    fs.existsSync( packagePackageJsonPath ) ? require( packagePackageJsonPath ) : {}
+  );
 
   const packageJson = Object.assign({}, packageJsonTemplate, {
-    name: `@blaze-elements/${packageName}`,
-    version: originalPackageJson.version || '1.0.0'
+    name: originalPackageJson.name || `@blaze-elements/${packageName}`,
+    version: originalPackageJson.version || '1.0.0',
+    dependencies: originalPackageJson.dependencies || undefined
```
Hotell, line 37: this is not necessary, if dependencies don't exist it will be `undefined` by default

toolbelt/generate-configs.js
```diff
@@ -3,29 +3,43 @@
...
   });
 
   const packageJsonString = (JSON.stringify(packageJson, null, 2) + "\n").replace(/packageName/g, packageName);
 
-  fs.writeFileSync( `./packages/${packageName}/package.json`, packageJsonString );
-
+  fs.writeFileSync( path.resolve(packageRootDir, 'package.json'), packageJsonString );
```
Hotell, line 42: next time maybe use just `resolve` without "long" object path chaining

```js
const {resolve} = require('path');

// .... 
```

Hotell, line 42: there is `'package.json'` string used at least 3 times ( also tsconfig ). Maybe rather assign it to variable and use that variable ( DRY, Refactoring friendly )

toolbelt/generate-configs.js
```diff
@@ -3,29 +3,43 @@
...
-  const originalPackageJson = require(`../packages/${packageName}/package.json`);
+  // Generate package.json file base on package package.json file and template
+  const packagePackageJsonPath = path.resolve(packageRootDir, 'package.json');
+
+  const originalPackageJson = (
+    fs.existsSync( packagePackageJsonPath ) ? require( packagePackageJsonPath ) : {}
+  );
 
   const packageJson = Object.assign({}, packageJsonTemplate, {
-    name: `@blaze-elements/${packageName}`,
```
Hotell, line 21: *suggestion* => we can maybe use power of desctructuring here:

```js
const {
  name=`@blaze-elements/${packageName}`,
  version='1.0.0',
  dependencies
} = fs.existsSync( packagePackageJsonPath ) 
  ? require( packagePackageJsonPath ) 
  : {}

const packageJson = {
 ...{name, version, dependencies}
};

savePackageConfigTo(packageRootDir, {name:packageName, config:packageJson});


function savePackageConfigTo(where,{name,config}={}){
  const configToWrite = serializeConfig(config).replace(/packageName/g, name);
  fs.writeFileSync( path.resolve(where, 'package.json'), configToWrite );
}

function serializeConfig( config ){
 return `${JSON.stringify( config, null, 2)}\n`;
}
```

Hotell (changes requested):
overall good job! 
just few nit picks about naming, simplifying and structuring code base.

in the future we should provide also tests for this

Hotell:
Interesting why did you just "closed"  this without further ado?

Issue #177, opened 2017-02-10
build: refactor to multiple targets - webpack config to be able to use externals


Hotell:
**Current behavior**

both index.min.js and index-with-deps.min.js are the same files with skatejs bundled in.

**Expected behavior**

we need to provide multiple targets webpack configs. https://webpack.js.org/concepts/configuration/#multiple-targets

### Things to do 

> ( Im putting it here because stupid github has hidden those comments )

this build entry points are not doing what you think they are doing.

`index.ts` imports `define` from skate, so as a end result. skate will be always bundled. This is not what we wanted.

What you need to do:

- [ ] for prod build you have to provide `externals` for main index.ts packages 
```
externals: {
  skatejs: 'skatejs'
}
```
- [ ] for prod build of `index-with-deps.ts` that vendor import is completely unnecessary and may increase bundle size. We are using three shaking, so only used functions/imports are bundled. only thing that you need to import there are `polyfills`

PR #188, opened 2017-02-13
Tweak build

### LINKED ISSUE #170: packaging: whole blaze-elements package should be under /packages
**Current behavior**

**Expected behavior**

We should have all source under `packages` directory to not mix up things and make structure of project chaotic. 

**Minimal reproduction of the problem with instructions**
What I would like to see is similar approach like material-components have https://github.com/material-components/material-components-web/tree/master/packages/material-components-web
So in the end everything is easily published via lerna

### LINKED ISSUE #180: build: toolbelt templates

devopsmariocom:
Closes #170
Closes #180


toolbelt/generate-configs.js
```diff
@@ -3,29 +3,43 @@
...
 
 console.log('Generating package.json and tsconfig.json for:');
 
+const packagesRootDir = path.resolve(__dirname, '..', 'packages');
+const templatesRootDir = path.resolve(__dirname, 'templates');
+
+const packageJsonTemplate = require( path.resolve(templatesRootDir, 'package.json') );
+const tsconfigJsonTemplate = require( path.resolve(templatesRootDir, 'tsconfig.json') );
+
 for (let packageName of packageList) {
```
hopkins-tk: This `for loop` is quite log.

It would be better to extract the content to a separate function and iterate through **forEach**.

toolbelt/generate-configs.js
```diff
@@ -3,29 +3,43 @@
...
 
   console.log(packageName);
 
-  fs.createReadStream('./toolbelt/templates/tsconfig.json').pipe(fs.createWriteStream(`./packages/${packageName}/tsconfig.json`));
+  const packageRootDir = path.resolve(packagesRootDir, packageName);
+
+  // Copy tsconfig.json template to each package folder
+  fs.writeFileSync(path.resolve(packageRootDir, 'tsconfig.json'), JSON.stringify(tsconfigJsonTemplate, null, 2) + "\n")
```
hopkins-tk: It might look better, when you extract it to a named function, so that everyone can see what is going on.

Also, why do we need **"\n"** at the end - is it really necessary?

toolbelt/generate-configs.js
```diff
@@ -3,29 +3,43 @@
...
 
-  const originalPackageJson = require(`../packages/${packageName}/package.json`);
+  // Generate package.json file base on package package.json file and template
+  const packagePackageJsonPath = path.resolve(packageRootDir, 'package.json');
+
+  const originalPackageJson = (
```
hopkins-tk: Following might be much more readable:

```
const originalPackageJson = fs.existsSync( packagePackageJsonPath )
  ? require( packagePackageJsonPath )
  : {};
```

toolbelt/generate-configs.js
```diff
@@ -3,29 +3,43 @@
...
 
   const packageJson = Object.assign({}, packageJsonTemplate, {
-    name: `@blaze-elements/${packageName}`,
-    version: originalPackageJson.version || '1.0.0'
+    name: originalPackageJson.name || `@blaze-elements/${packageName}`,
+    version: originalPackageJson.version || '1.0.0',
+    dependencies: originalPackageJson.dependencies || undefined
   });
 
   const packageJsonString = (JSON.stringify(packageJson, null, 2) + "\n").replace(/packageName/g, packageName);
```
hopkins-tk: Saving of the package JSON could be extracted to a separate function.

Also, why do we need **"\n"** at the end - is it really necessary?

toolbelt/generate-configs.js
```diff
@@ -3,29 +3,43 @@
...
+  const originalPackageJson = (
+    fs.existsSync( packagePackageJsonPath ) ? require( packagePackageJsonPath ) : {}
+  );
 
   const packageJson = Object.assign({}, packageJsonTemplate, {
-    name: `@blaze-elements/${packageName}`,
-    version: originalPackageJson.version || '1.0.0'
+    name: originalPackageJson.name || `@blaze-elements/${packageName}`,
+    version: originalPackageJson.version || '1.0.0',
+    dependencies: originalPackageJson.dependencies || undefined
```
hopkins-tk: Does explicit `undefined` make any difference?

toolbelt/generate-packages-list.js
```diff
@@ -1,18 +1,68 @@
...
+function generatePackageList( packageJson, packageList ) {
+  packageJson.packages = packageList;
+
+  fs.writeFileSync('./package.json', JSON.stringify(packageJson, null, 2) + "\n");
+}
+
+function generateIndex( filename, packageList, callback ) {
+  fs.writeFileSync(
+     filename,
+    `${packageList.map(callback).join(`\n`)}\n`
```
hopkins-tk, line 48: Is there any reason, why could't we use just:
```
packageList.map(callback).join(`\n`)
```
?

toolbelt/generate-packages-list.js
```diff
@@ -1,18 +1,68 @@
...
+        return accumulator;
+      }
+
+      accumulator[`@blaze-elements/${packageName}`] = packageJson.version;
+
+      return accumulator;
+    }, {}
+  );
+
+  fs.writeFileSync( path.resolve(blazeElementPackagePath, 'package.json'), JSON.stringify(blazeElementPackageJson, null, 2) + "\n" );
```
hopkins-tk: Again writing package JSON - I would suggest using of a newly created function - similar lines are throughout the code.

hopkins-tk:
Seems like it is solving several things.

Might be worth splitting to several PR:
- one for https://github.com/wc-catalogue/blaze-elements/issues/180
- one for barrel package/lerna publish


There is also opened issue https://github.com/wc-catalogue/blaze-elements/issues/177, which should be covered in a separate PR, but some of the changes might be related more to it...

Introduction of **blaze-elements** package is probable mostly related to https://github.com/wc-catalogue/blaze-elements/issues/170.

hopkins-tk (approved):

Issue #172, opened 2017-02-07
refactor(helpers): change definitions/events naming


Hotell:
**Current behavior**

event aliases are wrongly named `ClickEvent` etc

**Expected behavior**

they should be named `ClickHandler` as they are function types not event structure types

PR #192, opened 2017-02-14
refactor(helpers): Change event naming to splitted handler and event

davidrus:

```
[ ] Bugfix
[ ] Feature
[x] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Change naming #172 

**What is the new behavior?**
Event naming was splitted to separate handlers and events.

```
[ ] Yes
[x] No
```

If this PR contains a breaking change, please describe the impact and migration path for existing applications: ...

**Other information**:

### DIFF
--- a/packages/_helpers/definitions/events.ts
+++ b/packages/_helpers/definitions/events.ts
@@ -1,7 +1,11 @@
-  export type ClickEvent = typeof HTMLElement.prototype.onclick;
-  export type FocusEvent = typeof HTMLElement.prototype.onfocus;
-  export type BlurEvent = typeof HTMLElement.prototype.onblur;
-  export type KeyupEvent = typeof HTMLElement.prototype.onkeyup;
+export type ClickHandler = ( event: ClickEvent ) => any;
+export type ClickEvent = typeof HTMLElement.prototype.onclick;
+export type FocusHandler = ( event: FocusEvent ) => any;
+export type FocusEvent = typeof HTMLElement.prototype.onfocus;
+export type BlurHandler = ( event: BlurEvent ) => any;
+export type BlurEvent = typeof HTMLElement.prototype.onblur;
+export type KeyupHandler = ( event: KeyupEvent ) => any;
+export type KeyupEvent= typeof HTMLElement.prototype.onkeyup;
 
-  export type CustomChangeHandler<T> = ( event: CustomChangeEvent<T> ) => any;
-  export type CustomChangeEvent<T> = CustomEvent & { detail: { value: T } };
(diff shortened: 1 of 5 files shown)


packages/_helpers/definitions/events.ts
```diff
@@ -1,7 +1,11 @@
-  export type ClickEvent = typeof HTMLElement.prototype.onclick;
-  export type FocusEvent = typeof HTMLElement.prototype.onfocus;
-  export type BlurEvent = typeof HTMLElement.prototype.onblur;
-  export type KeyupEvent = typeof HTMLElement.prototype.onkeyup;
+export type ClickHandler = ( event: ClickEvent ) => any;
```
Hotell: We don't have to create `<Native>Event` because they are already within prototype definition `HTMLElement.prototype.<native>`.

`Event` types will be defined only for custom Events.

so it should look like this:

```ts
export type ClickHandler = typeof HTMLElement.prototype.onclick;
export type FocusHandler = typeof HTMLElement.prototype.onfocus;
export type BlurHandler = typeof HTMLElement.prototype.onblur;
export type KeyupHandler = typeof HTMLElement.prototype.onkeyup;

// custom events and handlers
export type CustomChangeHandler<T> = ( event: CustomChangeEvent<T> ) => any;
export type CustomChangeEvent<T> = CustomEvent & { detail: { value: T } };
```

packages/button/Button.test.tsx
```diff
@@ -3,6 +3,7 @@ import { h, mount } from 'bore';
 import { emit } from 'skatejs';
 
 import { Button } from './index';
+import { ClickEvent } from '../_helpers/definitions/events';
```
Hotell: we import from package root, not from private path

packages/button/Button.test.tsx
```diff
@@ -48,7 +49,7 @@ describe( Button.is, () => {
       it( `should handle click`, () => {
 
         let clickTriggered = false;
-        const handleClick = ( e: MouseEvent ) => { clickTriggered = true; };
+        const handleClick = ( e: ClickEvent ) => { clickTriggered = true; };
```
Hotell: this is rather wrong, pls see first review commit. `ClickEvent` is function in this case, so just remove it and use appropriate native event type ( in this case `MouseClickEvent` )

Hotell (changes requested):

Hotell:
just minor change, when resolved we can merge. thx!

davidrus:
Thank you @Hotell for review 👍

Hotell (approved):

PR #209, opened 2017-02-20
feat(toggle): add onChange emit when checked value changed

davidrus:
affects: @blaze-elements/toggle

- [x] Tests for the changes have been added (for bug fixes / features)

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Toggle does not emit onchange event when value changed

**What is the new behavior?**
OnChange was emitted when toggle checked was changed

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/toggle/Toggle.test.tsx
+++ b/packages/toggle/Toggle.test.tsx
@@ -1,17 +1,153 @@
-import { Toggle } from './Toggle';
-
-import { h, mount } from 'bore';
+import * as expect from 'expect';
+import { h, mount, WrappedNode } from 'bore';
+import { emit } from 'skatejs';
+import { GenericEvents } from '../_helpers';
+import { Toggle } from './index';
 
 describe( Toggle.is, () => {
 
-  it( 'should render', () => {
+  describe( `Custom element`, () => {
 
-    console.warn( 'Missing tests for Toggle!' );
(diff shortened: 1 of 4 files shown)


packages/toggle/Toggle.test.tsx
```diff
@@ -1,17 +1,153 @@
...
 
-    console.warn( 'Missing tests for Toggle!' );
+    } );
 
-    return mount(
-      <Toggle />
-    ).wait();
+    function checkColor( element: WrappedNode ) {
+      const label: Partial<HTMLInputElement> = element.one( 'label' ).node;
+      expect( label.getAttribute( 'class' ) ).toContain( 'warning' );
```
Hotell: just a nitpick. It's better to use appropriate api for specific task, in this case I would prefer `classList`

`expect(label.classList).toContain('warning')`
or 
`expect(label.classList.contains('warning')).toBe(true)`

Hotell (approved):
Just a little nitpick, it's not demanding.

But I would appreciate if you can incorporate it. thx

packages/toggle/Toggle.test.tsx
davidrus: First suggestion not working, I used second, thanks!!!

Issue #215, opened 2017-02-21
feat(select,option): implement bl-select with bl-option


Hotell:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => search github for a similar issue or PR before submitting
[x] feature request
[ ] support request => You can submit support request here, but we preffer to chat about it realtime at our [ngParty slack](https://ngparty.herokuapp.com/)
```

**Current behavior**
we don't support custom select element

**Expected behavior**

we need something like this https://material.angular.io/components/component/select

**POC:**

### Usage (within Preact):

```ts

const selectedState = 'CA';
const states = [{code: 'AL', name: 'Alabama'}...];

const App = () => (
<bl-select placeholder="Select state" value={selectedState} >
  { 
    states.map(state => {
       return ( 
          <bl-option value={state.code}>
            {option.name}
          </bl-option> 
       ) 
    }
   }
</bl-select>
)
```

- so active `bl-option` will be one, which is equal to `bl-select` `selected`
- we will use probably slot JS api for this as well https://developers.google.com/web/fundamentals/getting-started/primers/shadowdom#workwithslots

#### bl-select API
- @prop disabled: boolean
- @prop value: any
- @prop placeholder: string
- @prop open: boolean
- @event change: Event emitted when the selected value has been changed by the user.

#### bl-option API
- @prop disabled: boolean
- @prop value: any
- @event onSelect: Event emitted when the option is selected

#### Keyboard interaction:

- DOWN_ARROW: Focus next option
- UP_ARROW: Focus previous option
- ENTER or SPACE: Select focused item

##### Naive Implementation
```ts
type OptionProps = {}
@customElement('bl-option')
class Option extends Component<OptionProps>{
  
  @prop({type:Boolean}) disabled: boolean;
  @prop({type:Object}) value: any;   

  private _selected: boolean;
  get selected(){ return this._selected }
  private slot;

  get slotedValue(){
    return this.slot.assignedNodes({flatten: true});
  }
  get viewValue(){
    return this.slotedValue.textContent.trim();
  }

  select(){
    this._selected = true;
  }
  deselect(): void {
    this._selected = false;
  }
  renderCallback(){
     return (
       <div><ref={this.setSlot} slot/></div>
     )
  }
  
  @bind
  private setSlot(el: SlotHTMLElement ){
    this.slot = el;
  }
}

type SelectProps = {}
@customElement('bl-select')
class Select extends Component<SelectOptions>{
[...]

PR #221, opened 2017-02-22
feat(select): add Select and Option component

davidrus:

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

https://github.com/wc-catalogue/blaze-elements/issues/215
`<select>` element does not have blaze wc component

**What is the new behavior?**
- `<select>` has `<bl-select>` as blaze wc component
- `<option>` has `<bl-option>` as blaze wc component

**Example**
`
<bl-select value="en">
  <bl-option value="cs">Cesky<bl-option>
  <bl-option value="en">English<bl-option>
</bl-select>
`
- English is highlighted as selected
- Value of bl-select is actual selected option

```
[ ] Yes
[x] No
```


packages/select/Option.tsx
```diff
@@ -9,7 +9,7 @@ export type OptionProps = Props & EventHandlers;
 
 export type Props = {
   value?: string,
-}
+};
```
Hotell: these kind of commits has to use `style` instead of `chore`. It's just nit pick, I know, but we wanna stick to and follow our guidelines :)

if you tslint fix/ run formatter just for one package `style(package-name): apply formatter, fix lint errors`, if it's touching multiple packages just `style: apply formatter`

packages/select/Option.tsx
```diff
@@ -0,0 +1,63 @@
+import { h, Component } from 'skatejs';
+import { GenericEvents } from '../_helpers/index';
```
Hotell: multiple imports from `helpers` internals, pls move everything to import from one entry point ( `helpers` )

packages/select/Option.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+export type Events = {
+  click?: GenericEvents.ClickHandler,
+};
+export type EventHandlers = {
+  onClick?: GenericEvents.ClickHandler,
+};
+
+@shadyCssStyles()
+@customElement( 'bl-option' )
```
Hotell: As I showed you during grooming session, we register elements to customElements container solely in barrel file. Pls see `button` for reference

packages/select/Option.tsx
```diff
@@ -0,0 +1,63 @@
...
+export default class Option extends Component<OptionProps> {
+
+  @prop( { type: Boolean } ) disabled: boolean;
+  @prop( { type: Object } ) value: any;
+  @prop( { type: Boolean } ) _selected = false;
```
Hotell: underscores for class properties are used only if you create getter/setter to that property ( which you do )
pls explicitly use private `@prop( { type: Boolean } ) private _selected: false`,

packages/select/Option.tsx
```diff
@@ -0,0 +1,63 @@
...
+    this._selected = true;
+  }
+
+  deselect(): void {
+    this._selected = false;
+  }
+
+  renderCallback() {
+
+    return <SelectCardItem selected={this.selected}>
```
Hotell: use parenthesis and move code to new line
```
return (
   <SelectCardItem>.....</>
)
```

packages/select/Select.tsx
```diff
@@ -0,0 +1,116 @@
+import { h, Component } from 'skatejs';
+import { GenericEvents, prop } from '../_helpers';
```
Hotell: multiple helpers import again from private api

packages/select/Select.tsx
```diff
@@ -0,0 +1,116 @@
...
+import Option from './Option';
```
Hotell: pls adhere to proper imports order
- 3rd party / node modules first
- other package import seconds
- internal imports last

packages/select/Select.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+export type Events = {
+  click?: GenericEvents.ClickHandler,
+};
+export type EventHandlers = {
+  onClick?: GenericEvents.ClickHandler,
+};
+
+@shadyCssStyles()
+@customElement( 'bl-select' )
```
Hotell: again pls see button, how we register custom elements to container

packages/select/Select.tsx
```diff
@@ -0,0 +1,116 @@
...
+export default class Select extends Component<SelectProps> {
+
+  @prop( { type: String } ) placeholder: string;
```
Hotell: this needst to be `attribute: { source: true } ` as well

packages/select/Select.tsx
```diff
@@ -0,0 +1,116 @@
...
+  @prop( { type: Boolean, attribute: { source: true } } ) open = true;
+  @prop( { attribute: true } ) value: any;
```
Hotell: do we really need 2 way reflection to attributes? what if value is object or so?

packages/select/Select.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  get selected(): Option {
+    return this._selected;
+  }
+
+  get css() {
+    return styles;
+  }
+
+  get selectedViewValue(): string {
```
Hotell: this is passing TSC? because it returns string or void, we should be consistent

packages/select/Select.tsx
```diff
@@ -0,0 +1,116 @@
...
+      this.open = false;
+    } );
+  }
+
+  renderCallback() {
+    return ( [
+
+      <SelectButton onClick={this.toggleOptions}>{this.selectedViewValue || this.placeholder} &#9660;</SelectButton>,
+
+      this.open && <SelectCard>
```
Hotell: 2 times open condition and parenthesis missing.

consider:

```ts
this.open && ...[
  <SelectCard>
     <slot ref={this.setSlot} slot="body" />
  </SelectCard>,
  <SelectOverlay onClick={this.toggleOptions} isFullpage isTransparent />
]
```

packages/select/Select.tsx
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  @bind
+  setSlot( element: HTMLSlotElement ) {
+    this.slotElement = element;
+  }
+
+  @bind
+  toggleOptions() {
+    this.open = !this.open;
```
Hotell: we dont directly set props, use `props` function

packages/select/components/Button.tsx
```diff
@@ -0,0 +1,7 @@
+// @FIXME this needs to be imoprted from 'bl-button' package, not relatively
+
+import { customElement } from '../../_helpers';
+import Button from '../../button/Button';
```
Hotell, line 4: `import Button from @baze-elements/button/Button`

packages/select/components/Card.tsx
```diff
@@ -0,0 +1,7 @@
+// @FIXME this needs to be imoprted from 'bl-button' package, not relatively
+
+import { customElement } from '../../_helpers';
+import { Card } from '../../card/Card';
```
Hotell, line 4: `import Card from @baze-elements/card/Card`

packages/select/components/CardItem.tsx
```diff
@@ -0,0 +1,7 @@
+// @FIXME this needs to be imoprted from 'bl-button' package, not relatively
+
+import { customElement } from '../../_helpers';
+import { CardItem } from '../../card/CardItem';
```
Hotell, line 4: `import CardItem from @baze-elements/card/CardItem`

packages/select/components/Overlay.tsx
```diff
@@ -0,0 +1,7 @@
+// @FIXME this needs to be imoprted from 'bl-button' package, not relatively
+
+import { customElement } from '../../_helpers';
+import { Overlay } from '../../overlay/Overlay';
```
Hotell, line 4: `import Overlay from @blaze-elements/overlay/Overlay`

packages/select/index.demo.tsx
```diff
@@ -0,0 +1,51 @@
+import { h, Component } from 'skatejs';
+import { customElement } from '../_helpers';
+import Select from './Select';
```
Hotell: once you register them correctly you will need to import from index
`import {Select,Options} from './index'`

packages/select/index.test.ts
```diff
@@ -0,0 +1 @@
+import './Button.test';
```
Hotell: uh? what is this?

packages/select/Select.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  private slotElement: HTMLSlotElement;
+  private _selected: Option;
+  private options: Option[];
+
+  get slottedValue(): any {
+    return this.slotElement.assignedNodes( { flatten: true } );
+  }
+
+  attributeChangedCallback() {
```
Hotell: whoa this is wrong, I guess you miss typed the life cycle hook? 
it should be`connectedCallback`

**NOTE**
 --> we never use `attributeChangedCallback` if we need custom watch for props we use `updatedCallback`

packages/select/Select.tsx
```diff
@@ -0,0 +1,116 @@
...
+    super.connectedCallback();
+    setTimeout(() => {
+      this.options = this.slottedValue;
+      this._setSelectionByValue( this.value );
+      this.open = false;
```
Hotell: rather use one method for updating `open`. leverage `toggleOpen` or rename that function to `setOpenTo(value)`

Hotell (changes requested):
Overall looking good, just few tweaks needed. thx

packages/select/Select.tsx
davidrus: I have some problems with Option import from ./index, I will talk to you later...

packages/select/Select.tsx
davidrus: Thats true...

packages/select/Select.tsx
davidrus: Yeah, but it shouldn't... :/

packages/select/components/Button.tsx
davidrus: I cannot do this before publishing helpers as module, because:
`Error:(9, 8) TS2307:Cannot find module '../_helpers'.` in Button.tsx 👎

davidrus: I keep TODO note here, and when we finish and publish helpers, we must refactor all packages where we using it.

packages/select/components/Card.tsx
davidrus: Same as button

packages/select/components/CardItem.tsx
davidrus: Same as button

packages/select/components/Overlay.tsx
davidrus: Same as button

packages/select/index.test.ts
davidrus: emmm... Just test for you as reviewer :D

packages/select/components/Button.tsx
Hotell: yup, but in the meantime pls apply what works => ( button )

Hotell (approved):

davidrus:
Thank You @Hotell for your time & gr8 review 🚶

Issue #236, opened 2017-03-05
fix(tag): input is set on wrong version


devopsmariocom:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => search github for a similar issue or PR before submitting
[ ] feature request
[ ] support request => You can submit support request here, but we preffer to chat about it realtime at our [ngParty slack](https://ngparty.herokuapp.com/)
```

**Current behavior**
tag has dependency on input ^1.0.0 which is outdated version and is breaking the build

detailed explanation:
- lerna download old npm package instead of linking
- package contain ts/tsx files which depends on _helpers folder as described #235

**Expected behavior**
we should have proper version

**Minimal reproduction of the problem with instructions**
`yarn test`

Hotell:
Pls always remember to have Closes #NumberOfTicket in your commits, otherwise related issue wont be closed automatically.

thx!

Hotell:
Closed by #237

PR #237, opened 2017-03-05
fix(tag): set input dependency to correct version

devopsmariocom:
- [x] Tests for the changes have been added (for bug fixes / features)

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Wrong version is set in tag/package.json which leads to errors on build
See #236 for details

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/blaze-elements/package.json
+++ b/packages/blaze-elements/package.json
@@ -33,7 +33,7 @@
     "@blaze-elements/collapsible": "^1.0.0",
     "@blaze-elements/common": "^1.0.0",
     "@blaze-elements/drawer": "^1.0.0",
-    "@blaze-elements/input": "^1.0.0",
+    "@blaze-elements/input": "^2.0.0",
     "@blaze-elements/modal": "^1.0.0",
     "@blaze-elements/nav": "^1.0.0",
     "@blaze-elements/overlay": "^1.0.0",
--- a/packages/tag/package.json
+++ b/packages/tag/package.json
@@ -22,6 +22,6 @@
   },
   "dependencies": {
     "@blaze-elements/common": "^1.0.0",
-    "@blaze-elements/input": "^1.0.0"
+    "@blaze-elements/input": "^2.0.0"
   }
 }


packages/blaze-elements/package.json
```diff
@@ -33,7 +33,7 @@
     "@blaze-elements/collapsible": "^1.0.0",
     "@blaze-elements/common": "^1.0.0",
     "@blaze-elements/drawer": "^1.0.0",
-    "@blaze-elements/input": "^1.0.0",
+    "@blaze-elements/input": "^2.0.0",
```
Hotell, line 36: we have 2.0 input? what was the BC ?

Hotell (approved):

Hotell (changes requested):
Hmm actually. this should be possible Breaking change for blaze-elements suite, because you are introducing new BC version for input.

Use case:

As a consumer I'm using `"blaze-elements":"1.2.3"` everything works. 

Now developer has bumped one package to new version because of breaking change. If I update `blaze-elements` to `1.2.4` ( which will be published according to your commit ), my app will brake - so it should be rather `"blaze-elements":"2.0.0"` 

OR!

is Lerna smart enhough to bump `blaze-elements` to `2.0.0` ?

packages/blaze-elements/package.json
devopsmariocom: I did release 2.0.1, because there was already 2.0.0 in npm, I don't see any related changes in git, so we most probably never find our.

devopsmariocom:
I didn't add `BREAKING CHANGE:` as there is not really any breaking change it's just fixing bad release as someone in the past did release version input 2.0.0 for unknown reason. If you want can add it there but not sure if it's really necessary.

packages/blaze-elements/package.json
Hotell: right, so this should be a breaking change in `blaze-elements` package

Hotell:
pls let's read what I've written one more time :) I think we are not on the same page

Hotell:
as discussed on internal status meeting, we can move forward as release will be resolved within another PR. thanks @elmariofredo

Hotell (approved):

PR #246, opened 2017-03-06
fix: remove @bind decorators from 'decko' because IE error

davidrus:
affects: @blaze-elements/calendar, @blaze-elements/select, @blaze-elements/tag

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

decko does not working in ie

**What is the new behavior?**
decko removed :)

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/calendar/index.demo.tsx
+++ b/packages/calendar/index.demo.tsx
@@ -1,5 +1,4 @@
 import { h, Component, prop } from 'skatejs';
-import { bind } from 'decko';
 import { GenericEvents, customElement } from '../_helpers';
 import { Calendar } from './index';
 
@@ -33,6 +32,11 @@ export class Demo extends Component<void> {
     }
   };
 
+  constructor() {
+    super();
+    this.dateChangeHandler = this.dateChangeHandler.bind( this );
+  }
+
   renderCallback() {
     return (
       <fieldset>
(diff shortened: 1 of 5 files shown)


Hotell:
This should fix just affected package which is select in this case, not other packages. Rather create issue for others. thx

Hotell:
anyway good job, although there is a opinion across the team that I'm constantly nitpicking and trying to "Educate" others in a Force manner. 

To prevent that please assign this to someone who can behave (according to team) and who will not have comments against your work. thx

hopkins-tk:
I think that we might be able to agree on following (we can discuss it tomorrow):
- ideally, commit should be about one thing - once you need to sue `and` to describe what it is about, it should be more commits
- from that it is clear, that it should be covering single package, thus should have single scope in most cases

However as this PR was considered more like remove `@bind` decorators, it has single thing it is about, even though cross several scopes.
Sure, could be multiple commits in single PR, but than we would end-up with something like this once squashing & merging.

We can start doing it correctly as we will have to if we will want to use scopes instead of affects - but currently, we use affects :disappointed:

hopkins-tk:
Replacement PR were created:
- https://github.com/wc-catalogue/blaze-elements/pull/249
- https://github.com/wc-catalogue/blaze-elements/pull/250

davidrus:
thank you @hopkins-tk for your PRs and @Hotell for comments. 
So I will close this PR to prefer #250 #249

